PR #70, opened 2025-11-26
Sync `package.json` dependencies when generating code

lorisleiva:
This PR adds 3 new options to the renderer which enable the `package.json` of the generated client to be created and/or updated such that its dependencies match what the generated code expect.

These options are:
- `packageFolder`: The path to the root of the package — i.e. where the `package.json` is/should be located. This option must be set for the `syncPackageJson` option below to work.
- `syncPackageJson`: Whether or not we should update the existing `package.json` if some of its dependencies are falling behind or create a new `package.json` if missing (default: `false`).
- `dependencyVersions`: A `Record<string, string>` such that keys are the name of external dependencies — e.g. `@solana/kit` — and the values are the version ranges we want to support — e.g. `^5.0.0`. The renderer offer default values for all external dependencies it relies on but this option may be used to override some of these values or add new ones.

Additionally, even when the `syncPackageJson` option is `false`, if the `packageFolder` is provided and there is an existing `package.json`, the renderer will go through its dependencies and check if any are missing or out of date for the generated code. A warning message will be logged if that's the case.

Note, when bumping the JS renderer to the next major version (v2), we can make sure the required path is not the path to the generated folder but the package folder instead. That way, we can make the `syncPackageJson` option default to `true`. This would mean `packageFolder` would be replaced by `generatedFolder` and would default to `src/generated`.


lorisleiva:
* **#70** <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/70?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/70?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94 (approved):
This looks great!

Issue #95, opened 2026-01-20
Generate a Account/Instruction -> JSON function for debugging


mcintyre94:
I think that a useful generated function for programs would be something that automates this structure (token as an example):

```ts
const instruction: TokenInstruction = identifyTokenInstruction(myTokenInstruction)

if (instruction === TokenInstruction.InitializeMint) {
  assertIsInstructionWithaccounts(myTokenInstruction)
  assertIsInstructionWithData(myTokenInstruction)
  return {
    instructionName: 'InitializeMint',
    parsed: parseInitializeMintInstruction(myTokenInstruction)
  }
}

if (instruction === TokenInstruction.InitializeAccount) {
  ...
}

...
```

This wouldn't be a very useful data type to work with as an application, because it can be any parsed instruction. But it would be extremely useful to drop in an eg `isTokenError` handler to quickly log the instruction that caused the error.

I think we could add this at the program level, and have a function like `getDebugInfoForInstruction(instruction: Instruction)` and `getDebugInfoForAccount(account: EncodedAccount)` that does something similar for accounts. For accounts we would use the `decodeMint` etc function, which returns an `Account` with decoded data.

It might also be useful to have data-only versions of these, so that if you just have an instruction/account data then you can quickly decode it regardless of its type. Again, probably only useful for debugging/developer tools, but I think it could be very useful.

PR #96, opened 2026-01-22
Generate a parseProgramInstruction function to identify + parse an instruction

mcintyre94:
This PR generates a new function in program clients that can be used to identify + parse an `Instruction & InstructionWithData` in one call.

It returns an object with `instructionType`, and the parsed instruction for that instruction type. I think this will be very useful for logging in error handlers:

```ts
if (isSystemError(cause, transactionMessage)) {
  const instruction = transactionMessage.instructions[cause.context.index]
  console.log(parseSystemProgramInstruction(instruction)) // instruction name, parsed accounts + decoded data
}
```

Part of ~#95~ #4

### DIFF
--- a/test/e2e/token/src/generated/programs/system.ts
+++ b/test/e2e/token/src/generated/programs/system.ts
@@ -47,6 +47,6 @@ export function parseSystemInstruction<TProgram extends string>(
             return { instructionType: SystemInstruction.CreateAccount, ...parseCreateAccountInstruction(instruction) };
         }
         default:
-            throw new Error('Unrecognized instruction type');
+            throw new Error('Unrecognized instruction type: ' + instructionType);
     }
 }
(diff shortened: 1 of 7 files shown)


test/e2e/token/src/generated/programs/system.ts
```diff
@@ -28,3 +36,17 @@ export function identifySystemInstruction(
 export type ParsedSystemInstruction<TProgram extends string = '11111111111111111111111111111111'> = {
     instructionType: SystemInstruction.CreateAccount;
 } & ParsedCreateAccountInstruction<TProgram>;
+
+export function debugSystemInstruction<TProgram extends string>(
+    instruction: Instruction<TProgram> & InstructionWithData<ReadonlyUint8Array>,
+): ParsedSystemInstruction<TProgram> {
```
mcintyre94: When I wrote the issue for this I didn't realise we had the `ParsedSystemInstruction` etc type already, which I think makes sense as the return type here.

Maybe the function should just be called `parseSystemInstruction` instead?

PR #109, opened 2026-02-04
Update to use Kit v6 for the renderer and defult packageJson

mcintyre94:
- Update all kit dependencies from `^5.0.0` to `^6.0.0`
- Use Kit `^6.0.0` for the default package.json
- Update tests (just needed to change from `BaseTransactionMessage` to `TransactionMessage`
- Unsure why this was affected, but there was a lint error for `Options` from `tsup` not being a type import. Just decided to include fixing that in this PR


lorisleiva (approved):
Awesome, thank you!

PR #113, opened 2026-02-12
Make `ImportMap` support subpath exports

lorisleiva:
This PR adds support for subpath exports in the `ImportMap` utility.

Previously, when resolving external dependencies, the `getExternalDependencies` function would return the full import path including subpath exports. This caused issues when checking dependencies since package.json lists root package names, not subpaths.

Now, subpath exports like `@solana/kit/program-client-core` are correctly resolved to their root package name (`@solana/kit`) when determining external dependencies.

### DIFF
--- a/.changeset/icy-dancers-train.md
+++ b/.changeset/icy-dancers-train.md
@@ -1,0 +1,5 @@
+---
+'@codama/renderers-js': patch
+---
+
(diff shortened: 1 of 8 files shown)


mcintyre94 (approved):
Nice!

PR #115, opened 2026-02-12
Use `SolanaErrors` for generated errors

lorisleiva:
This PR updates generated program clients to use `SolanaError` from `@solana/kit` instead of throwing generic `Error` objects. This improves error handling by providing specific error codes and structured context.

### DIFF
--- a/src/fragments/discriminatorCondition.ts
+++ b/src/fragments/discriminatorCondition.ts
@@ -91,7 +91,6 @@ function getFieldConditionFragment(
 ): Fragment {
     const field = scope.struct.fields.find(f => f.name === discriminator.name);
     if (!field || !field.defaultValue) {
-        // TODO: Coded error.
         throw new Error(
             `Field discriminator "${discriminator.name}" does not have a matching argument with default value.`,
         );
--- a/.changeset/strong-breads-spend.md
+++ b/.changeset/strong-breads-spend.md
@@ -1,0 +1,5 @@
+---
+'@codama/renderers-js': minor
+---
+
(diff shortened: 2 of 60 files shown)


src/fragments/discriminatorCondition.ts
```diff
@@ -91,7 +91,6 @@ function getFieldConditionFragment(
 ): Fragment {
     const field = scope.struct.fields.find(f => f.name === discriminator.name);
     if (!field || !field.defaultValue) {
-        // TODO: Coded error.
         throw new Error(
             `Field discriminator "${discriminator.name}" does not have a matching argument with default value.`,
         );
```
mcintyre94: Is the removed comment still valid, since this one isn't a `SolanaError`?

mcintyre94 (approved):
Awesome! Really glad we have these as proper coded errors now!

src/fragments/discriminatorCondition.ts
lorisleiva: Yeah I did this because this is not a generated error but one that gets thrown when running the renderer directly. There's quite a few of these all over the renderer and none of them have this comment so I decided to unify them by removing it.

If we wanted coded errors here, we'd need to add them in the @codama/errors package instead which is a different issue now that renderers are out of the monorepo.

I think for now non-coded errors for these scenarios are fine though.

lorisleiva:
### Merge activity

* **Feb 17, 11:19 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/codama-idl/renderers-js/115).

PR #117, opened 2026-02-13
Replace `useGranularImports` option with `kitImportStrategy`

lorisleiva:
This PR replaces the `useGranularImports` boolean option with a new `kitImportStrategy` option that provides three strategies for resolving `@solana/kit` imports in generated code:

- `'granular'` — Always import from standalone packages (e.g. `@solana/addresses`, `@solana/codecs-strings`).
- `'preferRoot'` (default) — Import from `@solana/kit` when a symbol is available from the root entrypoint, otherwise fall back to granular packages.
- `'rootOnly'` — Only import from `@solana/kit`, using subpath exports (e.g. `@solana/kit/program-client-core`) when needed.

With the new default `'preferRoot'` strategy, symbols exported from the `@solana/kit` root entrypoint are imported from `@solana/kit`, while symbols only available from subpackages (e.g. `@solana/program-client-core`) fall back to their granular package instead of using subpath exports.

The `'rootOnly'` strategy is useful when installing `@solana/kit` as a `peerDependency`, since it ensures all imports resolve through a single package. However, because it relies on subpath exports (e.g. `@solana/kit/program-client-core`), it may require setting TypeScript's `moduleResolution` to `"bundler"` for these imports to resolve correctly.


mcintyre94 (approved):
Nice, great docs!

lorisleiva:
### Merge activity

* **Feb 17, 11:19 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/codama-idl/renderers-js/117).

PR #126, opened 2026-02-22
Add export for options from utils

piliugin-anton:
Add export for missing RenderOptions type


piliugin-anton:
> Hi there, happy to merge this. Could you just fix the lint issue (ordering I think)?
> 
> Out of curiosity, what makes you need these option types?

I'm making a developers tool, need to access all renderers dynamically, so I typed it, only JavaScript renderer was missing an options export

PR #129, opened 2026-03-18
Add PDA derivation to generated program plugin

amilz:
## Summary
- Adds `pdas` field to plugin type and factory function, following the existing accounts/instructions pattern
- PDA find functions are passed through directly (no runtime wrapper needed since they're stateless)
- Programs with only PDAs now generate a valid plugin with `object` requirements type
- Regenerated e2e fixtures (associated token program now includes `pdas` field)

## Test plan
- [x] 8 new unit tests covering: PDA types, PDA objects, import verification, PDA-only programs, no-PDAs negative assertion, instructions+PDAs combo, requirements for PDA-only
- [x] All 447 unit tests pass across node/browser/react-native
- [x] All e2e tests pass
- [x] TypeScript compilation clean

Ref: https://x.com/redacted_noah/status/2034008748130505087

### DIFF
--- a/.changeset/add-pdas-to-plugin.md
+++ b/.changeset/add-pdas-to-plugin.md
@@ -1,0 +1,5 @@
+---
+'@codama/renderers-js': minor
+---
+
+Add PDA derivation to generated program plugin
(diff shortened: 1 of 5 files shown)


lorisleiva (approved):
Nice, great addition! Thank you!

EDIT: Sounds like you just need to run `pnpm lint:fix` and we should be good to merge.

amilz:
Pushed the format cleanup!

PR #133, opened 2026-04-09
Enable isolatedDeclarations in renderer and generated output

ioxde:
Adds explicit return type annotations to satisfy [`isolatedDeclarations`](https://www.typescriptlang.org/tsconfig/#isolatedDeclarations) in both the renderer itself and generated output.

`isolatedDeclarations` lets bundlers emit `.d.ts` without full type inference. Generated packages that enable it downstream would error on discriminator helpers and encoder functions missing return types.


src/fragments/programPlugin.ts
```diff
@@ -200,7 +200,7 @@ function getProgramPluginFunctionFragment(
     );
 
     return fragment`export function ${programPluginFunction}() {
-    return <T extends ${programPluginRequirementsType}>(client: T) => {
+    return <T extends ${programPluginRequirementsType}>(client: T): T & { ${programPluginKey}: ${programPluginType} } => {
         return { ...client, ${programPluginKey}: <${programPluginType}>{ ${fields} } };
```
lorisleiva: I think because we now hint the return type, we no longer need to hint the inner object (to double check).


```suggestion
    return <T extends ${programPluginRequirementsType}>(client: T): T & { ${programPluginKey}: ${programPluginType} } => {
       return { ...client, ${programPluginKey}: { ${fields} } };
```

src/visitors/getTypeManifestVisitor.ts
```diff
@@ -52,7 +52,57 @@ export function getTypeManifestVisitor(input: {
...
     stack?: NodeStack;
-}) {
+}): Visitor<
+    TypeManifest,
+    | 'accountNode'
+    | 'amountTypeNode'
+    | 'arrayTypeNode'
+    | 'arrayValueNode'
+    | 'booleanTypeNode'
+    | 'booleanValueNode'
```
lorisleiva: You can simplify that list by using the `RegisteredTypeNode` and `RegisteredValueNode` unions.

test/e2e/system/test/_setup.ts
```diff
@@ -92,7 +94,7 @@ export const signAndSendTransaction = async (
   return signature;
 };
 
-export const getBalance = async (client: Client, address: Address) =>
+export const getBalance = async (client: Client, address: Address): Promise<bigint> =>
```
lorisleiva:
```suggestion
export const getBalance = async (client: Client, address: Address): Promise<Lamports> =>
```

test/e2e/token/test/_setup.ts
```diff
@@ -95,7 +97,7 @@ export const signAndSendTransaction = async (
   return signature;
 };
 
-export const getBalance = async (client: Client, address: Address) =>
+export const getBalance = async (client: Client, address: Address): Promise<bigint> =>
```
lorisleiva:
```suggestion
export const getBalance = async (client: Client, address: Address): Promise<Lamports> =>
```

src/fragments/discriminatorConstants.ts
```diff
@@ -78,21 +79,34 @@ export function getFieldDiscriminatorConstantFragment(
...
 
     const name = camelCase(`${prefix}_${discriminatorNode.name}`);
-    const encoder = visit(field.type, typeManifestVisitor).encoder;
+    const typeManifest = visit(field.type, typeManifestVisitor);
+    const encoder = typeManifest.encoder;
     const value = visit(field.defaultValue, typeManifestVisitor).value;
-    return getConstantFragment({ ...scope, encoder, name, value });
+    const needsTypeAnnotation =
+        isNode(field.defaultValue, 'bytesValueNode') ||
+        (isNode(field.defaultValue, 'constantValueNode') && isNode(field.defaultValue.value, 'bytesValueNode'));
```
lorisleiva: I think we need to accept more scenarios here than just `BytesValueNodes` because it could be that a discriminator has a complex type that needs type annotation. If I understand correctly, this `needsTypeAnnotation` check is to prevent things like `number` or `boolean` to be explicitly typed because then TypeScript would complain that it's unnecessary. If that's true, then we could simply check a few scalar types against `typeManifest.strictType` to know if it's needed or not. Something like:

```ts
const valueType = typeManifest.strictType;
const needsTypeAnnotation = !["string", "number", "boolean"].contains(valueType);
```

It would be nice to add a few more tests to ensure these cases work as expected.

lorisleiva (changes requested):
Thanks! I think this is net value for client maintainers that want to quickly generate type definitions. I've made a few comments but I think we're almost there.

src/fragments/discriminatorConstants.ts
ioxde: > TypeScript would complain that it's unnecessary

Typescript wouldn't do this, but eslint might with [@typescript-eslint/no-inferrable-types](https://typescript-eslint.io/rules/no-inferrable-types/) enabled. I think linting should be disabled on generated code.

The check is so `const x = 42` will be inferred as `42` instead of `number` and still pass with `isolatedDeclarations` enabled.

After implementing your suggestion and adding `bigint` to this list, I found `bigint` discriminators were dropping the `n` suffix (`7` instead of `7n`). This latest change added them back- let me know if that isn't something you want changed and I'll revert it.

src/visitors/getTypeManifestVisitor.ts
ioxde: b4c543ae46588a760597f75b92e788cfbb335358 is that what you were thinking?

src/fragments/programPlugin.ts
ioxde: Removed and verified with the e2e tests.

src/visitors/getTypeManifestVisitor.ts
lorisleiva: Exactly, thanks!

src/fragments/discriminatorConstants.ts
```diff
@@ -78,21 +83,38 @@ export function getFieldDiscriminatorConstantFragment(
...
+        typeManifest.strictType,
+    );
+    return getConstantFragment({ ...scope, encoder, name, value, valueType });
+}
+
+function resolveDiscriminatorValue(rawValue: Fragment, isNumberValue: boolean, strictType: Fragment) {
+    let value = rawValue;
+    if (strictType.content === 'bigint' && isNumberValue) {
+        value = Object.freeze({ ...value, content: `${value.content}n` });
+    }
```
lorisleiva: nit: You can just compose fragments like so:

```suggestion
    const value = strictType.content === 'bigint' && isNumberValue
      ? fragment`${rawValue}n`
      : rawValue;
```

src/fragments/discriminatorConstants.ts
lorisleiva: Good catch! I've just made a nit comment but looks good to me.

ioxde:
Do you want me to squash all of the commits into one?

lorisleiva (approved):

Issue #127, opened 2026-02-25
Add option to generate clients with useEsmExtensions


Woody4618:
Currently when generating a codama js client for a module with 

    "target": "es2020",
    "module": "nodenext",

you get the following warnings: 

```
Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.ts(2834)
```

resulting in errors: 

```
Module '"../clients/js/src/generated/index.js"' has no exported member 'letMeBuyProgram'.ts(2305)
```

Would be great for that case to have the option generate the .js extensions 

```

export * from "./accounts/index.js";
export * from "./errors/index.js";
export * from "./instructions/index.js";
export * from "./programs/index.js";
export * from "./types/index.js";
```

instead of 

```
export * from "./accounts";
export * from "./errors";
export * from "./instructions";
export * from "./programs";
export * from "./types";
```

lorisleiva:
Implemented in #177, #178 and #179

PR #134, opened 2026-04-09
Emit nodenext-compatible imports with explicit .js extensions

ioxde:
Emit imports with explicit `.js` extensions in generated code.

Generated code now uses explicit `.js` extensions on all relative imports and sets `module: "nodenext"` / `moduleResolution: "nodenext"` in the generated `tsconfig.json`.

[`nodenext`](https://www.typescriptlang.org/tsconfig/#moduleResolution) requires explicit extensions and proper ESM semantics, producing output that works in Node.js, bundlers, and other runtimes. TypeScript 6 deprecates the legacy node10 resolution, and [TypeScript 7](https://github.com/microsoft/typescript-go) [removes it entirely](https://github.com/microsoft/TypeScript/issues/62200).

#127


lorisleiva:
See https://github.com/codama-idl/renderers-js/pull/135#issuecomment-4212329156

PR #135, opened 2026-04-09
Update Typescript to v6

ioxde:
Depends on #134

Update to Typescript `v6`

Sources:
https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/
https://gist.github.com/privatenumber/3d2e80da28f84ee30b77d53e1693378f


lorisleiva:
Sorry but the `.js` and `/index.js` ESM-only extensions is not what we're looking to achieve here by default. In practice, program client maintainers are going to want to expose their packages as ESM and CJS builds which is best done as CJS first to avoid TypeScript issues. That being said, I'd be okay accepting an option for this that defaults to keeping export paths the way they are.

Correct me if I'm wrong but TS v6 doesn't have anything to do with this change so I'll close this PR. Feel free to open a separate one for v6 only and another that makes this ESM-only approach opt-in.

ioxde:
`nodenext` + `"type": "commonjs"` + `verbatimModuleSyntax` was the issue here, not `.js`. I should have figured that out before putting the combined PRs in, sorry about that.

PR #136, opened 2026-04-09
Use bracket notation for index signature property access

ioxde:
#2 

I ran into this issue and solved it using the post-processor below without this fix:

```ts
export function fixNodeEnvAccess(dir: string): void {
	for (const entry of readdirSync(dir)) {
		const fullPath = join(dir, entry);
		if (statSync(fullPath).isDirectory()) {
			fixNodeEnvAccess(fullPath);
			continue;
		}
		if (!entry.endsWith('.ts')) continue;

		let content = readFileSync(fullPath, 'utf8');
		if (!content.includes('process.env.NODE_ENV')) continue;
		content = content.replaceAll(
			'process.env.NODE_ENV',
			"process.env['NODE_ENV']",
		);
		writeFileSync(fullPath, content);
	}
}
```

### DIFF
--- a/.changeset/bracket-notation-index-signature.md
+++ b/.changeset/bracket-notation-index-signature.md
@@ -1,0 +1,5 @@
+---
+"@codama/renderers-js": patch
+---
+
+Use bracket notation for index signature property access
--- a/src/fragments/errorPage.ts
+++ b/src/fragments/errorPage.ts
@@ -58,7 +58,7 @@ function getErrorMessagesFragment(scope: Pick<RenderScope, 'nameApi'> & { progra
     );
 
     return fragment`let ${mapName}: Record<${errorUnionType}, string> | undefined;
-if (process.env.NODE_ENV !== 'production') {
+if (process.env['NODE_ENV'] !== 'production') {
   ${mapName} = { ${messageEntries} };
 }`;
 }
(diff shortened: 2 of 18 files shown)


lorisleiva:
@trevor-cortex

trevor-cortex (approved):
## Summary

Fixes generated code to be compatible with TypeScript's `noPropertyAccessFromIndexSignature` rule by using bracket notation for `process.env['NODE_ENV']` in the error page fragment, and for `env['npm_package_version']` in the tsup configs. Also enables the rule in the root `tsconfig.json` and all 5 e2e tsconfigs so the check is enforced in CI going forward. Regenerated e2e error fixtures are updated accordingly.

## Key things to watch out for

- **Rule enforcement scope**: The root `tsconfig.json` covers `src`, `test`, and `tsup.config.ts` but excludes `test/e2e`. Each e2e project gets its own tsconfig update, so generated fixture code will now be type-checked with the rule enabled — this is the important feedback loop that will catch any future regressions.
- **Completeness**: The fix in `src/fragments/errorPage.ts` is the only fragment source change. This is consistent with the diffs in the regenerated e2e fixtures, which only touch the two `process.env.NODE_ENV` sites per generated errors file. No other generator sites appear to need updating based on the e2e regeneration delta.
- **tsup config**: `env.npm_package_version` → `env['npm_package_version']` is correct because `env` is imported from `node:process`, which exposes an index signature.

- Worth confirming that `pnpm test:unit` doesn't have any assertions hard-coding the old dot-notation string (`process.env.NODE_ENV`) in fragment snapshot tests. I couldn't enumerate the test directory, but if CI is green, any such assertion has already been updated or simply doesn't exist.
- Worth running `pnpm test:e2e` locally or verifying CI ran it — the e2e projects now have `noPropertyAccessFromIndexSignature: true`, so if any other generator output still uses dot notation on an index signature type, type-check would fail there.
- The `.changeset` file correctly marks this as a `patch` bump, which is appropriate — it's a purely mechanical change to the emitted code that doesn't alter runtime behavior (bracket vs. dot access on `process.env.NODE_ENV` is semantically identical at runtime and equally well preserved by bundlers doing dead code elimination).

LGTM.

lorisleiva (approved):
Thanks!

PR #137, opened 2026-04-10
Upgrade TypeScript to v6

ioxde:
Trimmed down version of #135 with only the necessary changes for project-level types. Generated code might need to be tested in a followup commit.

`"moduleResolution": "node"` is [deprecated in 6.0](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#deprecated---moduleresolution-node-aka---moduleresolution-node10) and [removed in 7.0](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#preparing-for-typescript-70).

```js
tsconfig.declarations.json:2:5 - error TS5107: Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0. Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
  Visit https://aka.ms/ts6 for migration information.

2     "compilerOptions": {
      ~~~~~~~~~~~~~~~~~

Found 1 error in tsconfig.declarations.json:2

 ELIFECYCLE  Command failed with exit code 2.
 ```

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -62,17 +62,21 @@
...
         "@solana/prettier-config-solana": "0.0.5",
         "@types/node": "^25",
         "@types/semver": "^7.7.1",
         "agadoo": "^3.0.0",
-        "eslint": "^9.35.0",
+        "eslint": "^9.39.1",
+        "eslint-plugin-jest": "^29.2.1",
+        "eslint-plugin-react-hooks": "^7.0.1",
+        "globals": "^16.5.0",
+        "typescript-eslint": "^8.49.0",
```
lorisleiva: What's this for?

tsconfig.declarations.json
```diff
@@ -3,7 +3,8 @@
         "declaration": true,
         "declarationMap": true,
         "emitDeclarationOnly": true,
-        "outDir": "./dist/types"
+        "outDir": "./dist/types",
+        "rootDir": "./src"
```
lorisleiva, line 7: Will this change the export paths in the `dist` folder?

tsconfig.json
```diff
@@ -4,20 +4,21 @@
         "composite": false,
         "declaration": true,
         "declarationMap": true,
-        "esModuleInterop": true,
```
lorisleiva, line 7: Why no longer needed?

tsconfig.json
```diff
@@ -4,20 +4,21 @@
...
         "forceConsistentCasingInFileNames": true,
         "inlineSources": false,
         "isolatedModules": true,
         "lib": [],
-        "moduleResolution": "node",
+        "module": "nodenext",
+        "moduleResolution": "nodenext",
```
lorisleiva, line 12: Would it makes sense to drop them altogether? I believe the new default in the `bundle` mode (TBC) which seems like a great default.

lorisleiva (changes requested):

lorisleiva:
@trevor-cortex Can you please review this PR and explain to me why (or if) the changes are important in the context of a TypeScript v6 upgrade (whilst keeping v7 in mind).

tsconfig.json
```diff
@@ -4,20 +4,21 @@
...
         "noFallthroughCasesInSwitch": true,
         "noPropertyAccessFromIndexSignature": true,
         "noUnusedLocals": true,
         "noUnusedParameters": true,
         "preserveWatchOutput": true,
         "skipLibCheck": true,
```
trevor-cortex, line 18: 👍 `nodenext`/`nodenext` is the right choice for a library targeting Node.js. Worth noting that `bundler` would have been the other option (and is actually the new TS 6.0 default when no module is specified), but `nodenext` is more correct here since this project emits declarations and needs strict ESM resolution semantics.

tsconfig.declarations.json
```diff
@@ -3,7 +3,8 @@
         "declaration": true,
         "declarationMap": true,
         "emitDeclarationOnly": true,
-        "outDir": "./dist/types"
+        "outDir": "./dist/types",
```
trevor-cortex, line 6: This is specifically needed because TS 6.0 changed `rootDir` to default to the tsconfig directory (`.`) instead of inferring from input files. Without it, declarations would nest under `dist/types/src/` instead of `dist/types/`. Note this only matters here because this config actually emits (via `emitDeclarationOnly`) — the base `tsconfig.json` doesn't need it since it's used for type-checking only.

tsconfig.json
```diff
@@ -4,20 +4,21 @@
...
         "strict": true,
-        "target": "ESNext"
+        "target": "ESNext",
+        "types": ["node"]
     },
```
trevor-cortex, line 22: Required in TS 6.0 — the `types` default changed from "all @types" to `[]`. Without this, you'd lose `@types/node` globals (`process`, `Buffer`, `__dirname`, etc).

trevor-cortex (approved):
## Summary

This PR upgrades TypeScript from v5.9 to v6.0 and updates the ESLint toolchain to match. The tsconfig changes are minimal and well-targeted. Here's a breakdown of each change and why it matters:

### `tsconfig.json` changes

**1. `moduleResolution: "node"` → `moduleResolution: "nodenext"` (+ `module: "nodenext"`)**

This is the critical fix. `moduleResolution: "node"` (aka `node10`) is **deprecated in TS 6.0** and will be **removed in TS 7.0**. It encoded Node.js 10-era resolution behaviour and doesn't reflect how modern Node resolves modules (e.g. exports maps, ESM support). `nodenext` is the correct choice for a library that targets Node.js directly (as opposed to `bundler` for bundled apps). Adding explicit `module: "nodenext"` is the right pairing.

**2. Removed `esModuleInterop: true`**

In TS 6.0, `esModuleInterop` defaults to `true` and setting it to `false` is deprecated. Since it was already `true` here, removing it is just config cleanup — no behavioural change.

**3. Added `"types": ["node"]`**

TS 6.0 changed the `types` default from "all @types packages" to an **empty array**. Without this explicit setting, `@types/node` wouldn't be picked up anymore, causing errors like `Cannot find name 'process'`. This is a required migration step.

**4. `target` stays `"ESNext"`**

No change needed here since the project was already on `ESNext` (the new default is `es2025`, but `ESNext` >= `es2025` so no issue).

### `tsconfig.declarations.json` change

**5. Added `rootDir: "./src"`**

TS 6.0 changed the `rootDir` default from "inferred common directory of input files" to the **tsconfig.json directory** (i.e. `.`). Without this explicit `rootDir`, declaration files would emit to `dist/types/src/index.d.ts` instead of `dist/types/index.d.ts`. Since this config emits declarations with `outDir: "./dist/types"`, setting `rootDir: "./src"` preserves the expected output structure. Good that it's only on the declarations config since the base `tsconfig.json` doesn't emit.

### Dependency updates

**6. `@solana/eslint-config-solana` v5 → v6 + new peer deps**

The eslint config package bumped its peer requirements to match TS 6.0 compatibility. The new direct deps (`eslint-plugin-jest`, `eslint-plugin-react-hooks`, `globals`, `typescript-eslint`) were previously transitive or provided by the config — now they need to be explicit peers. This is just keeping the ESLint toolchain compatible with TS 6.0.

### TS 7.0 readiness

All the deprecated options that were addressed here (moduleResolution node10, implicit rootDir, implicit types) will become **hard errors** in TS 7.0 (the Go-based rewrite). This PR properly addresses them rather than using `ignoreDeprecations: "6.0"`, which is the right call.

---

Looks clean. The tsconfig changes are the minimum necessary set for TS 6.0 compatibility, and they're all forward-compatible with 7.0. The PR description mentions generated code may need testing in a followup — that's the right scope boundary for this.

tsconfig.json
lorisleiva: @trevor-cortex Why is `nodenext` more correct than `bundler`? Isn't the latter the new default btw?

trevor-cortex: Good question — I should nuance my earlier comment. `bundler` is indeed the new TS 6.0 default (when no `module` is specified, it resolves to `es2022` which defaults `moduleResolution` to `bundler`).

Both `nodenext` and `bundler` are valid here, but `nodenext` is the safer choice for this project specifically because of **declaration files**. This project uses `tsup` to bundle the JS output but uses `tsc` to emit individual `.d.ts` files (via `tsconfig.declarations.json`). That's the key detail.

The TypeScript docs explicitly warn about this pattern: if you use `moduleResolution: "bundler"` with a bundler for JS but `tsc` for declarations, the `.d.ts` files may contain import specifiers that are valid for a bundler but **invalid in Node.js** (e.g. missing file extensions). Consumers of this library who use `moduleResolution: "nodenext"` would then get errors on those declaration files.

Since `nodenext` enforces stricter resolution rules (like requiring file extensions on relative imports), it ensures the declaration files this project emits are compatible with the widest range of consumer configurations — both `bundler` and `nodenext` users can consume them safely.

That said, if the source already uses extensionless imports everywhere and `nodenext` would require a bunch of changes to add `.js` extensions, `bundler` would be the pragmatic choice. But if the project compiles clean with `nodenext` (which it appears to, since the PR doesn't touch any source imports), it's the more correct option for a published library.

package.json
lorisleiva: @trevor-cortex Is any of this required in the scope of this PR?

tsconfig.declarations.json
lorisleiva: @trevor-cortex Can you check the build steps to make sure we don't need to change any paths in the `package.json`.

tsconfig.json
lorisleiva: @trevor-cortex Wdyt?

tsconfig.declarations.json
trevor-cortex: Checked the build setup — no path changes needed in `package.json`.

The build step is `tsup && tsc -p ./tsconfig.declarations.json`. The `tsc` invocation uses `include: ["src/index.ts", "src/types"]` with `outDir: "./dist/types"`. 

Without the new `rootDir: "./src"`, TS 6.0 would default `rootDir` to the tsconfig directory (`.`), which would produce `dist/types/src/index.d.ts`. With `rootDir: "./src"` set explicitly, it produces `dist/types/index.d.ts` — which is exactly what `package.json` expects:

```
"types": "./dist/types/index.d.ts"
"exports": { "types": "./dist/types/index.d.ts", ... }
```

So this `rootDir` addition is actually *preserving* the existing output structure. The paths in `package.json` don't need to change — they were correct before and remain correct with this fix.

tsconfig.json
trevor-cortex: The ESLint config bump (`@solana/eslint-config-solana` v5 → v6) is likely required because v5 may not be compatible with TS 6.0 (e.g. the `typescript-eslint` parser needs to support TS 6.0 syntax). However, the new explicit peer deps (`eslint-plugin-jest`, `eslint-plugin-react-hooks`, `globals`, `typescript-eslint`) are arguably a separate concern — they're satisfying peerDependency requirements of the new eslint config version, not directly related to the TS upgrade itself.

If you want to keep this PR minimal (just the tsconfig + TS version bump), you could potentially split the ESLint config upgrade into a separate PR. But practically, if `pnpm lint` fails without the eslint config bump, then it's all part of the same "make the project work on TS 6.0" scope.

package.json
trevor-cortex: Good question. The `@solana/eslint-config-solana` bump from v5 to v6 is **not strictly required** for the TypeScript v6 upgrade itself. The TS 6.0 changes (tsconfig updates) are independent of the ESLint config version.

However, it's likely included because the v5 eslint config may have compatibility issues with TS 6.0 (e.g. `typescript-eslint` needing a version that supports TS 6.0's new module resolution semantics). If `pnpm lint` passes on the `main` branch with TS 6.0 + the old eslint config v5, then this bump could be split into a separate PR.

The new explicit deps (`eslint-plugin-jest`, `eslint-plugin-react-hooks`, `globals`, `typescript-eslint`) are peer dependencies of `@solana/eslint-config-solana` v6 — with pnpm's strict dependency resolution, they need to be listed explicitly. They wouldn't be needed if you stayed on v5.

So: if you want to keep this PR minimal (just TS 6.0 compat), you could try keeping `@solana/eslint-config-solana` at v5 and only bump it if linting breaks. That would remove all 6 of these dependency changes and leave just the `typescript` version bump.

trevor-cortex:
Re: `rootDir` in `tsconfig.declarations.json` — I checked the build steps and export paths. No `package.json` changes needed. ✅

The build script is `rimraf dist && tsup && tsc -p ./tsconfig.declarations.json`. The `tsup` step handles JS bundles (unaffected). The `tsc` step emits declarations to `./dist/types` via `outDir`.

With explicit `rootDir: "./src"`, declarations still land at `./dist/types/index.d.ts` — matching both:
- `"exports" → "types": "./dist/types/index.d.ts"`
- `"types": "./dist/types/index.d.ts"`

Without this change, TS 6.0's new default (`rootDir` = tsconfig directory = `.`) would emit to `./dist/types/src/index.d.ts` instead, breaking those paths. So `rootDir: "./src"` just preserves the current behaviour.

trevor-cortex:
Re: removing `esModuleInterop` — In TS 6.0, `esModuleInterop` now **defaults to `true`** (previously defaulted to `false`). Since this config already had it set to `true`, removing it has zero behavioural change — it's just removing a now-redundant setting.

Additionally, explicitly setting `esModuleInterop: false` is deprecated in TS 6.0 and will warn. So removing it entirely is the cleanest move — aligns with the new defaults and avoids any future confusion.

tsconfig.json
lorisleiva: @trevor-cortex "if the source already uses extensionless imports everywhere" > it does. We are not using any of these `.js` and `/index.js` extensions. We aim to export for both CJS and ESM whilst making TS happy on both side.

trevor-cortex: Ah, that's a really important detail I should have checked — but it actually still works fine here because the project has `"type": "commonjs"` in `package.json`.

Under `nodenext`, TypeScript follows Node.js's actual resolution rules, which differ between CJS and ESM:
- **CJS files** (`.ts` in a `"type": "commonjs"` package): extensionless imports are perfectly valid — Node's `require()` has always resolved them.
- **ESM files** (`.mts` or `.ts` in a `"type": "module"` package): file extensions are mandatory — that's where `nodenext` would force you to add `.js` extensions.

Since this project is `"type": "commonjs"`, all `.ts` files are treated as CJS by `nodenext`, so extensionless imports like `'./utils/importMap'` work without any changes. The dual CJS/ESM output is handled by `tsup` (the bundler), not `tsc`, so the module format of the built output is independent of what `tsc` sees during type-checking.

This also means the `.d.ts` files emitted by `tsc` will have extensionless specifiers, which is fine — they're consumed alongside the `exports` map in `package.json`, which explicitly points consumers to the right files regardless of format.

So `nodenext` is both correct and compatible here — no source changes needed.

package.json
lorisleiva: @trevor-cortex Hmm why would bumping `@solana/eslint-config-solana` from v5 to v6 make use install these internal dependencies? Can you double check your findings?

trevor-cortex: You're right to push back — I was wrong about these being "internal dependencies that now need to be explicit." 

Looking at the lockfile, `@solana/eslint-config-solana` v6.0.0 declares all of these as **peer dependencies**:

- `eslint-plugin-jest: ^29.2.1`
- `eslint-plugin-react-hooks: ^7.0.1`
- `globals: ^16.5.0`
- `typescript-eslint: ^8.49.0`

(along with `@eslint/js`, `eslint`, `eslint-plugin-simple-import-sort`, `eslint-plugin-sort-keys-fix`, `eslint-plugin-typescript-sort-keys`, `jest`, and `typescript`)

With pnpm's strict dependency model, peer deps need to be explicitly listed by the consumer. But this isn't a change from v5 → v6 that *introduced* these as new peer deps. More likely, v5 had a similar set of peer deps, and the previous `package.json` was relying on pnpm's `autoInstallPeers: true` (which is set in the lockfile) to resolve them transitively without listing them.

So the real question is: **why are these 4 now listed explicitly when they weren't before?** It could be that v5 had them as regular dependencies (not peer deps), or that v6 added them as new peer deps. Without checking v5's `package.json` I can't say for certain, but my earlier claim that this was about pnpm strictness was an oversimplification. 

Bottom line: if you want to test whether the eslint config bump is needed at all, try keeping v5 and running `pnpm lint` with TS 6.0. If it passes, you can drop all 6 dependency changes from this PR.

tsconfig.json
lorisleiva: Another comment from Trevor below mentioned `true` is the new default in v6.

package.json
lorisleiva: @ioxde Sorry for the AI noise but I think these should go. Lmk if I'm missing something.

tsconfig.json
lorisleiva: Trevor answered my question

package.json
ioxde: Done! No you didn't miss anything. I reverted the unnecessary package updates. Do you want me to merge the commits into one and force push?

ioxde: After reinstalling with the d52663e50b9a4d9b342ce3783ba1b4cc5ed1c550 change eslint complained about being unsupported. I pushed 8726025cbd960708eb765f3f6ecae46793693fdd as the minimal amount of changes I could make to get rid of that warning.

```sh
pnpm lint

> @codama/renderers-js@2.1.0
> eslint . && prettier --check .

=============


WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.


* @typescript-eslint/typescript-estree version: 8.43.0
* Supported TypeScript versions: >=4.8.4 <6.0.0
* Your TypeScript version: 6.0.2


Please only submit bug reports when using the officially supported version.


=============
Checking formatting...
All matched files use Prettier code style!
```

After 8726025cbd960708eb765f3f6ecae46793693fdd: 
```sh
pnpm lint

> @codama/renderers-js@2.1.0 lint
> eslint . && prettier --check .

Checking formatting...
All matched files use Prettier code style!
```

See also:
https://github.com/anza-xyz/eslint-config-solana/commit/22456fd25aee9040985efb534705abf86224912c
https://github.com/typescript-eslint/typescript-eslint/commit/493341709fe7d1d6433332b7bd2724e3332c7cdf

lorisleiva:
Closing in favour of https://github.com/codama-idl/renderers-js/pull/202

PR #138, opened 2026-04-10
Emit .js extension in generated files

### LINKED ISSUE #127: Add option to generate clients with useEsmExtensions

ioxde:
Trimmed down version of #134 without the `nodenext` changes. 

Adding `.js` is required with certain typescript configurations / ESM modules. This change only adds support for all import patterns.

Nice LLM generated table:

| environment | without `.js` | with `.js` |
  |---|---|---|
  | `moduleResolution: "node"` (legacy) | works | works |
  | `moduleResolution: "bundler"` | works | works |
  | `moduleResolution: "node16"` | fails | works |
  | `moduleResolution: "nodenext"` | fails | works |
  | CommonJS (`require()`) | works | works |
  
  Closes #127


lorisleiva:
Hmm as [discussed here](https://github.com/codama-idl/renderers-js/pull/135#issuecomment-4212329156):

> In practice, program client maintainers are going to want to expose their packages as ESM and CJS builds which is best done as CJS first to avoid TypeScript issues. That being said, I'd be okay **accepting an option** for this that defaults to **keeping export paths the way they are**.

This is not what this PR does so I'll close but feel free to open a new one that does what we discussed.

ioxde:
Sorry I meant to address that in an edit but didn't get around to it. I can add an option but I think it should be default enabled. `.js` is what is [supported by both CJS and ESM](https://github.com/codama-idl/renderers-js/pull/137#discussion_r3073217939). Excluding the `.js` actually makes the generated typescript less compatible with various `tsconfig.json` configurations, while adding `.js` makes it compatible with (almost) all* (that I am aware of) configurations.

It is [required for any consumer that is using `tsc` to emit](https://www.typescriptlang.org/tsconfig/#allowImportingTsExtensions) their files for use with ESM.

More typescript references: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html
https://nodejs.org/api/esm.html#mandatory-file-extensions

\* I was able to find one usage example that `.js` is **NOT** the correct pattern, and that's [Node's native TypeScript execution](https://nodejs.org/api/typescript.html#determining-module-system) which requires `.ts` extensions (extensionless and `.js` do not work). They love throwing a wrench in the ecosystem 😂

If you'd still rather have it off by default let me know, I can do a followup PR no problem.

PR #139, opened 2026-04-12
Kill e2e test validator on exit and wait for startup

ioxde:
I noticed after running `pnpm test:e2e` solana-test-validator would never exit. This change will shutdown the test validator if started by the tests. If there is already a test validator running it will re-use it and leave it running.

Screenshot 2026-04-12 at 9 45 36 AM

### DIFF
--- a/test/e2e/test.sh
+++ b/test/e2e/test.sh
@@ -1,11 +1,42 @@
 #!/usr/bin/env bash
 set -eux
 
-function start_validator() {
-    if ! lsof -t -i:8899; then
-        echo "Starting solana-test-validator"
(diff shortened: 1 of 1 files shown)


lorisleiva:
I'd rather move toward litesvm in the near future (currently working on Kit plugins for this). This will remove the need for a local validator altogether.

ioxde:
[Is litesvm the correct choice for `e2e` tests?](https://litesvm.github.io/litesvm/tutorial.html#when-should-i-use-solana-test-validator) Either way, isn't this a good fix until litesvm is integrated? It took a bit of time trying to figure out why my laptop battery was draining

PR #140, opened 2026-04-14
Use extendClient in program plugins and bump @solana/kit to ^6.4.0

lorisleiva:
This updates the program plugin code generator to use `extendClient` from `@solana/plugin-core` instead of manually spreading the client object. The return type now uses `Omit<T, key> & { key: Plugin }` instead of `T & { key: Plugin }`, which allows proper type narrowing when extending clients with overlapping plugin keys. The minimum `@solana/kit` version is bumped to `^6.4.0` across all default dependency versions and e2e test packages.

### DIFF
--- a/.changeset/stale-lies-stare.md
+++ b/.changeset/stale-lies-stare.md
@@ -1,0 +1,5 @@
+---
+'@codama/renderers-js': minor
+---
+
+Use `extendClient` from `@solana/plugin-core` in generated program plugins instead of manually spreading the client object. This improves type narrowing when composing plugins with overlapping keys. Bump minimum `@solana/kit` dependency to `^6.4.0`.
--- a/src/fragments/programPlugin.ts
+++ b/src/fragments/programPlugin.ts
@@ -189,6 +189,7 @@ function getProgramPluginFunctionFragment(
     const programPluginType = nameApi.programPluginType(programNode.name);
     const programPluginRequirementsType = nameApi.programPluginRequirementsType(programNode.name);
     const programPluginKey = nameApi.programPluginKey(programNode.name);
+    const extendClient = use('extendClient', 'solanaPluginCore');
 
     const fields = mergeFragments(
         [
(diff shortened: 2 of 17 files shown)


lorisleiva:
* **#142** <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/142?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#140** <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/140?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/140?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

trevor-cortex (approved):
## Summary

This PR updates the program plugin code generator to use `extendClient` from `@solana/plugin-core` instead of manually spreading the client object (`{ ...client, key: { ... } }`). The return type changes from `T & { key: Plugin }` to `Omit<T, "key"> & { key: Plugin }`, which enables proper type narrowing when composing plugins with overlapping keys. All `@solana/*` default dependency versions are bumped from `^6.1.0` to `^6.4.0`.

## Review

Clean, mechanical change. The core logic update is in `src/fragments/programPlugin.ts` — a single new `use('extendClient', 'solanaPluginCore')` call and two template literal changes. The import mapping in `importMap.ts` already maps `solanaPluginCore` → `@solana/kit` (default) / `@solana/plugin-core` (granular), so the generated import will resolve correctly in both strategies.

The `Omit<T, key> & { key: Plugin }` pattern is the right approach — it replaces any existing property with the same key rather than creating an impossible intersection type (`T & { key: Plugin }` where `T` already has a different `key`).

All e2e generated files are consistent with the new template output, and the test expectation is updated to match.

- The `<PluginType>` type assertion on the object literal (e.g. `<SplTokenPlugin>{ ... }`) is worth noting — this is a deliberate cast. Since the object is being constructed inline with dynamic helpers like `addSelfFetchFunctions`, TypeScript may not be able to fully verify the shape matches the plugin type. This was likely already the case implicitly before; now it's explicit.
- The `^6.4.0` bump applies to all 14 `@solana/*` default dependencies uniformly. If `extendClient` was introduced in a specific minor version, it might be worth confirming that `6.4.0` is indeed the minimum version that exports it.
- The lockfile diff is large (~780 lines) but is just the natural result of bumping `@solana/kit` in the e2e test packages.

mcintyre94 (approved):
Nice, thankyou!

lorisleiva:
### Merge activity

* **Apr 20, 10:40 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/codama-idl/renderers-js/140).
* **Apr 20, 10:40 AM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.com/github/pr/codama-idl/renderers-js/140).

PR #141, opened 2026-04-15
Add event codegen support

ioxde:
NOTE: I don't like throwing out AI generated PRs but I think in this instance it is the right move. I made the typescript and rust version at the same time. Close if this is not the direction you want to go in.

## What this does

Adds event codegen support following the same fragment-based architecture as instructions and accounts.

**Per-event files** (`events/{name}.ts`) — event type, decoder, discriminator constants, and a `decode{Name}()` convenience wrapper. Events are decoder-only since they're emitted by programs, not constructed by clients.

**Program-level utilities** — `{Program}Event` enum, `identify{Program}Event()`, `parse{Program}Event()`, and a `Parsed{Program}Event` discriminated union type.

**Plugin integration** — adds an `events` field to the existing `{Program}Plugin` type (alongside `accounts`, `instructions`, `pdas`) with a `{Program}PluginEvents` type and decoder instances.

**Key design decision:** Events without constant discriminators in a `hiddenPrefixTypeNode` get individual files but are excluded from program-level aggregation (enum/identify/parse) — there's no reliable way to identify them at runtime.

## Limitations / Blocking

- [Uses `throw new Error(...)` for identification/parse failures](https://github.com/codama-idl/renderers-js/pull/141/changes/bcdaea075853bd79a6573c4ec0760a027a0f691e#r3088146127) probably should switch to `SolanaError` once event-specific error codes land in `@solana/errors`
- Only constant discriminators are validated in the decode wrapper (not field discriminators)
- [`*events ?? []` pattern](https://github.com/codama-idl/renderers-js/pull/141/changes/bcdaea075853bd79a6573c4ec0760a027a0f691e#diff-4c538efe2a1538d3eae073dd09c153b393d74cf499f1850961c670b948dc8d22R203) that should probably be fixed upstream

 ## Tests

- Event rendering across all discriminator variants (constant, field, none)
- Program-level enum/identify/parse generation
- Plugin type generation with event decoders
- E2e fixtures for Raydium CPMM (`LpChangeEvent`, `SwapEvent`)


src/visitors/getRenderMapVisitor.ts
```diff
@@ -170,11 +199,14 @@ export function getRenderMapVisitor(
                         leavesOnly: !renderScope.renderParentInstructions,
                     }).filter(isNotInternal);
                     const definedTypesToExport = getAllDefinedTypes(node).filter(isNotInternal);
+                    // program.events may be undefined when no events exist.
+                    const eventsToExport = getAllEvents(node).filter(Boolean).filter(isNotInternal);
```
ioxde, line 203: This guard and the `*events ?? []` patterns should probably be fixed upstream so they can be removed.

See https://github.com/codama-idl/codama/pull/991

src/fragments/programEvents.ts
```diff
@@ -0,0 +1,154 @@
...
+        c => c.join('\n'),
+    );
+
+    const readonlyUint8Array = use('type ReadonlyUint8Array', 'solanaCodecsCore');
+
+    return fragment`export function ${programEventsIdentifierFunction}(event: { data: ${readonlyUint8Array} } | ${readonlyUint8Array}): ${programEventsEnum} {
+    const data = 'data' in event ? event.data : event;
+    ${discriminatorsFragment}
+    // TODO: Use SolanaError once event-specific error codes are added to @solana/errors.
+    throw new Error('The provided event data does not match any known ${programNode.name} event.');
```
ioxde: This is most likely a blocker to this PR that will need to be addressed upstream [anza-xyz/kit/packages/errors](https://github.com/anza-xyz/kit/tree/main/packages/errors)

PR #143, opened 2026-05-01
feat: expose identify/parse helpers on generated program plugins

amilz:
## Summary
- Re-expose `identifyAccount`, `identifyInstruction`, and `parseInstruction` on the generated program plugin, so consumers can call `client.myProgram.identifyInstruction(ix)` without importing the per-function helpers.
- Emission mirrors the existing rules for those helpers (account/instruction discriminators, respecting `renderParentInstructions`).
- Adds fragment, page, and e2e (`dummy`) round-trip coverage; e2e fixtures regenerated.

## Test Plan
- `pnpm test:unit`
- `cd test/e2e/dummy && pnpm test`

### DIFF
--- a/.changeset/yellow-rice-smoke.md
+++ b/.changeset/yellow-rice-smoke.md
@@ -1,0 +1,5 @@
+---
+'@codama/renderers-js': minor
+---
+
+Expose `identifyAccount`, `identifyInstruction`, and `parseInstruction` on generated program plugins. When a program has accounts or instructions with discriminators, the plugin object now surfaces the corresponding identifier and parser helpers as `client.myProgram.identifyAccount(...)`, `client.myProgram.identifyInstruction(...)`, and `client.myProgram.parseInstruction(...)`, making it easier to build indexers without re-importing the per-function helpers.
--- a/src/fragments/programPlugin.ts
+++ b/src/fragments/programPlugin.ts
@@ -1,5 +1,6 @@
 import {
     CamelCaseString,
+    getAllInstructionsWithSubs,
     InstructionAccountNode,
     InstructionArgumentNode,
     InstructionNode,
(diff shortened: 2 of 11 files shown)


lorisleiva:
@trevor-cortex

trevor-cortex (approved):
## Summary

Re-exposes `identifyAccount`, `identifyInstruction`, and `parseInstruction` on generated program plugins so consumers can call them off the client (e.g. `client.myProgram.parseInstruction(ix)`) without importing the per-program helpers separately. Useful for indexers and anything that switches on instruction kind.

The plugin type and the plugin function are both updated symmetrically, and emission is gated by the same conditions that drive the underlying helper generation:

- `identifyAccount` is emitted iff at least one account has a discriminator (mirrors `getProgramAccountsIdentifierFunctionFragment`).
- `identifyInstruction` / `parseInstruction` are emitted iff at least one instruction (after the same `getAllInstructionsWithSubs({ leavesOnly: !renderParentInstructions })` walk used by `getProgramInstructionsFragment`) has a discriminator.

Threading `renderParentInstructions` into `getProgramPluginFragment`/`getProgramPluginTypeFragment`/`getProgramPluginFunctionFragment` is required for that second condition to stay in sync — `programPage.ts` already had it in scope, so the plumbing is local.

## Things to watch

- **Condition parity**: `hasAccountIdentifier` and `hasInstructionIdentifier` duplicate the predicates from `programAccounts.ts` / `programInstructions.ts`. They match today; if those upstream conditions ever change (e.g. additional gating beyond `discriminators.length > 0`), these will need to change in lockstep. A short comment is already present on `hasInstructionIdentifier`; worth keeping in mind for future edits.
- **Field ordering**: the new keys are appended after `pdas` in both the type and the function. Cosmetic only, but placing `identifyAccount` next to `accounts` and `identifyInstruction`/`parseInstruction` next to `instructions` would read more naturally. Not blocking — current placement is consistent across type + function and matches the regenerated e2e fixtures.

- E2E fixtures across `dummy`, `system`, `token`, `associatedToken`, and `wenTransferGuard` have been regenerated and look consistent with the new emission rules (e.g. `associatedToken` only gets `identifyInstruction`/`parseInstruction` because it has no accounts; `token` gets all three; `dummy` gets only the instruction pair).
- The new `dummy` AVA test exercises the plugin via a stub `{} as DummyPluginRequirements` cast — fine here because `identifyInstruction`/`parseInstruction` are bare references that don't read from the client, but worth remembering if either helper ever becomes client-aware.
- Tests cover the four interesting cases: discriminators present on both sides, sub-instruction-only discriminators, no discriminators (negative case), and `renderParentInstructions: true` flipping a parent-only discriminator into emission. Good coverage.
- Changeset is `minor` and the description is accurate.

lorisleiva (approved):
Perfection, Thank you!

PR #146, opened 2026-06-17
Support @solana/kit 6.10.0 in generated program plugins

lorisleiva:
This PR makes the JavaScript renderer compatible with `@solana/kit` 6.10.0, whose `extendClient` now returns the homomorphic `ExtendedClient<T, TAdditions>` type instead of a plain `Omit<T, keyof TAdditions> & TAdditions` intersection. The generated `xxxProgram()` factory used to hard-code that intersection as its return annotation, which is no longer assignable to what `extendClient` returns, so generated clients fail to type-check. The factory now annotates its return type with `ExtendedClient` directly, and the default `@solana/*` dependency versions written into a generated client's `package.json` via `syncPackageJson` are bumped to `^6.10.0` so the new type is actually available. The renderer's own `@codama/*` dependencies are also bumped to the latest published versions.


lorisleiva:
* **#146** <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/146?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/146?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #145, opened 2026-05-06
Add event generation from IDL


lognorman20:
## Context

Our Anchor IDL includes `events` entries and matching event payload structs in `types`. Codama parses those events into event nodes, but the JS renderer does not currently emit a typed event decoding surface. Consumers using Codama-generated TypeScript clients still need custom repo-specific glue to decode program logs into typed events.

## Problem

`@codama/renderers-js` generates accounts, instructions, PDAs, programs, errors, and defined types, but does not emit Anchor event helpers from `idl.events`. This forces downstream projects to manually generate event enums, discriminator maps, log parsing, and payload decoding despite that information already existing in the IDL.

## Request

Add first-class event generation to the JS renderer for Anchor IDL events. Generated output should include typed event payloads, discriminators, event identification, event decoding from raw bytes, and optional parsing from Solana `Program data: <base64>` logs.

shadeglare:
Agreed. Also the scalar constants should also be generated. Not a great pleasure to sync such stuff manually.

lorisleiva:
Implemented in #174

PR #174, opened 2026-08-01
feat: add events generation

ChiefWoods:
This PR adds first-class event generation to the JavaScript renderer, resolving https://github.com/codama-idl/renderers-js/issues/145.

Each `eventNode` is rendered into a dedicated `events/` page containing its discriminator constants, a typed event payload, encoder/decoder/codec functions and a `parseXxxEvent` helper that validates the event discriminators — constant, field and size discriminators are all supported — before decoding. Program pages additionally expose an event enum and an `identifyMyProgramEvent` helper mirroring the existing account and instruction identification helpers.

Event codecs reference their discriminator constants instead of inlining the bytes, thanks to new event awareness in the type manifest visitor, and all generated names can be customised through the new `eventDataType`, `eventParseFunction`, `programEventsEnum`, `programEventsEnumVariant` and `programEventsIdentifierFunction` name transformers. The anchor E2E fixture gains a `GuardCreated` event proving the full `nodes-from-anchor` conversion, compilation and runtime round-trip.

Two notes: Solana Kit does not yet define event-specific error codes, so the generated event helpers throw plain `Error`s for now, and parsing events out of `Program data:` log strings is intentionally left for a follow-up.


lorisleiva:
@trevor-cortex

src/fragments/discriminatorConstants.ts
```diff
@@ -45,6 +45,21 @@ export function getDiscriminatorConstantFragment(
     }
 }
 
+/**
+ * Derives the camel-case name of the constant rendered for a constant discriminator.
+ *
+ * The first constant discriminator of a node gets no index suffix, whilst subsequent
+ * ones are numbered from 2 — e.g. `myEventDiscriminator` and `myEventDiscriminator_2`.
```
trevor-cortex: Docblock nit: since the whole string goes through `camelCase(...)`, the `_2` suffix collapses — the actual output is `myEventDiscriminator2`, not `myEventDiscriminator_2` (consistent with the `COMPOSITE_EVENT_DISCRIMINATOR2` expectation in `test/eventsPage.test.ts`). Worth fixing the example so it matches reality.

src/visitors/getTypeManifestVisitor.ts
```diff
@@ -77,6 +80,26 @@ export function getTypeManifestVisitor(input: {
...
+        if (!eventNode) return null;
+        const constantDiscriminators = (eventNode.discriminators ?? []).filter(
+            isNodeFilter('constantDiscriminatorNode'),
+        );
+        const serializedConstant = JSON.stringify(constant);
+        const index = constantDiscriminators.findIndex(
+            discriminator => JSON.stringify(discriminator.constant) === serializedConstant,
+        );
+        if (index < 0) return null;
+        const prefix = camelCase(nameApi.eventDataType(eventNode.name));
```
trevor-cortex: This prefix derivation (`camelCase(nameApi.eventDataType(eventNode.name))`) must stay in lockstep with the `prefix: camelCase(eventDataName)` passed to `getDiscriminatorConstantsFragment` in `eventPage.ts` — if they ever diverge, the codec would reference a constant that was declared under a different name, and nothing would catch it at generation time. Consider extracting a tiny shared helper (e.g. `getEventDiscriminatorPrefix(nameApi, eventName)`) next to `getConstantDiscriminatorName` so the declaration and the reference can't drift apart.

src/fragments/eventPage.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    const resolvedData = resolveNestedTypeNode(node.data);
+    const condition = getDiscriminatorConditionFragment({
+        ...scope,
+        dataName: 'data',
+        discriminators,
+        ifTrue: `return ${decoderFunction}().decode(data);`,
+        programNode: findProgramNodeFromPath(scope.eventPath)!,
+        struct: isNode(resolvedData, 'structTypeNode') ? resolvedData : structTypeNode([]),
```
trevor-cortex, line 94: Question on the `structTypeNode([])` fallback: if an event has a non-struct payload (e.g. the enum case in the tests) *and* a `fieldDiscriminatorNode`, the condition fragment would be asked to resolve a field against an empty struct. Does `getDiscriminatorConditionFragment` fail with a clear error there, or silently generate a broken condition? It's a nonsensical input, so a generation-time failure is fine — just checking the failure mode isn't cryptic generated code. (Same fallback pattern is used in `programEvents.ts`.)

trevor-cortex (approved):
## Summary

This PR adds first-class event generation: each `eventNode` gets a dedicated `events/` page (discriminator constants, typed payload, encoder/decoder/codec, and a `parseXxxEvent` helper validating constant/field/size discriminators), and program pages gain an event enum plus an `identifyXxxEvent` helper mirroring the account/instruction equivalents. The type manifest visitor becomes event-aware so event codecs reference their discriminator constants instead of inlining bytes, five new name transformers cover all generated names, and the anchor E2E fixture proves the full round-trip. Docs (`README`, `CLAUDE.md`/`AGENTS.md`) and the changeset (minor, correctly) are all in place.

The implementation follows the repo's established patterns closely — fragments compose subfragments, naming goes through `nameApi`, discriminator logic reuses `getDiscriminatorConditionFragment`/`getDiscriminatorConstantsFragment`, and test coverage is thorough (constant/field/size discriminators, offsets, multiple discriminators, non-struct payloads, internal nodes, custom name transformers, empty-program barrel omission).

## Things to watch

- **Constant-name derivation is duplicated across declaration and reference sites.** The event page derives the constant prefix as `camelCase(eventDataName)` while the type manifest visitor independently derives `camelCase(nameApi.eventDataType(eventNode.name))`. They agree today, but if either drifts the generated code would reference a non-existent constant. See inline comment.
- **Program-page identify helper inlines bytes while event pages reference constants.** `identifyWenTransferGuardEvent` re-encodes `new Uint8Array([...])` through `fixEncoderSize(getBytesEncoder(), 8)` on every call, whereas the event page uses `containsBytes(data, GUARD_CREATED_EVENT_DISCRIMINATOR, 0)`. This mirrors the existing account identify helpers, so it's consistent — but flagging in case importing the constants from `generatedEvents` was intended as part of this PR's "reference constants instead of inlining" goal, or as a follow-up.

- For a program mixing discriminated and non-discriminated events, the event enum contains all variants but the identifier function can only ever return the discriminated ones (and throws otherwise). This seems like the right trade-off and matches the `heartbeat` test, just worth being aware of.
- The `JSON.stringify` structural matching in `getEventDiscriminatorConstantName` is key-order-sensitive, but since nodes come from the `@codama/nodes` factories this is fine in practice.
- The plain `Error` throws in the generated helpers are called out as intentional in the PR description (no Kit event error codes yet) — no action needed.
- The README gains an intro paragraph but the Options section wasn't extended for the five new `nameTransformers` keys; if that section enumerates keys elsewhere, a follow-up tweak may be warranted.

src/fragments/eventPage.ts
lorisleiva: Good question, it fails loudly at generation time (`getFieldConditionFragment` throws `Field discriminator "key" does not have a matching argument with default value.`) rather than emitting broken code, and 843da98 now locks that behaviour in with a unit test.

trevor-cortex (approved):
## Re-review

All three points from my previous review have been addressed — nice follow-through:

1. **Constant-name drift** — resolved. `discriminatorConstants.ts` now exports `getConstantDiscriminatorName` and a shared `getEventDiscriminatorPrefix` helper, and both the event page (declaration site) and the type manifest visitor's `getEventDiscriminatorConstantName` (reference site) derive the prefix through it. The two can no longer drift silently.
2. **Docblock nit** — resolved. The `getConstantDiscriminatorName` docblock now correctly shows `myEventDiscriminator2` and explains why the underscore collapses.
3. **`structTypeNode([])` fallback question** — answered with a test. The new `it fails loudly when a field discriminator has no matching payload field` test proves generation fails at render time with a clear error (`Field discriminator "key" does not have a matching argument with default value.`) rather than emitting broken code. Exactly the behaviour I was hoping for.

One tiny residual nit (non-blocking, see inline): the field-discriminator branch in `getTypeManifestVisitor.ts` still derives the event prefix inline via `camelCase(nameApi.eventDataType(...))` instead of calling the new `getEventDiscriminatorPrefix` helper. Same expression, so it's correct today — just the one remaining spot the helper was created to eliminate.

Everything else stands from the first review: the program-page identify helper inlining bytes remains a reasonable follow-up candidate, and the plain `Error` throws are intentional per the PR description. Still approved.

trevor-cortex:
Following up on the nit from the re-review body — the exact spot is in `src/visitors/getTypeManifestVisitor.ts`, in the field-discriminator handling near the bottom of the visitor (~L832):

```ts
const discriminatorPrefix =
    discriminatorParent && isNode(discriminatorParent, 'eventNode')
        ? camelCase(nameApi.eventDataType(discriminatorParent.name))
        : discriminatorParent?.name;
```

The event branch could call `getEventDiscriminatorPrefix(nameApi, discriminatorParent.name)` instead of repeating the expression inline — it's the one remaining site the shared helper doesn't cover. Purely cosmetic, fine to leave for a follow-up.

trevor-cortex (approved):
## Re-review (3rd pass)

The residual nit from my last review is now resolved: the field-discriminator branch in `getTypeManifestVisitor.ts` no longer derives the event prefix inline. It's been refactored into a single `discriminatorParent = instructionNode ?? accountNode ?? eventNode` resolution, with the prefix going through the shared `getEventDiscriminatorPrefix(nameApi, ...)` helper for events and falling back to the node name for instructions/accounts. That was the last spot outside the helper, so constant-name derivation is now fully centralised — declaration site (event page), constant-value reference site (`getEventDiscriminatorConstantName`), and field-discriminator default-value site all go through the same code path.

The refactor also reads cleaner than the previous ternary chains, and the precedence order (instruction > account > event) matches the path-based lookup semantics used elsewhere in the visitor.

Everything else stands from prior reviews:

- Shared prefix helper prevents silent drift between constants and their references. ✅
- Field discriminators on non-struct payloads fail loudly at render time, with a test proving it. ✅
- The program-page `identifyXxxEvent` helper still inlines discriminator bytes rather than importing the event constants — consistent with the account/instruction identify helpers, fine as a follow-up.
- Plain `Error` throws are intentional per the PR description (no Kit event error codes yet).

Nothing new to flag. Ship it. 🚀

lorisleiva (approved):

lorisleiva:
Thank you!

PR #175, opened 2026-08-01
feat: add constants generation

ChiefWoods:
`program.constants` now gets included in client generation, rendering a `constants.ts` that follows Borsh type serialization.


ChiefWoods:
Fixed e2e generated files that were missed in earlier commits

ChiefWoods:
Bumped codama dependencies to include the fix from https://github.com/codama-idl/codama/pull/1030

lorisleiva:
Thanks! I've got a few PRs to merge from different authors that will all create merge conflicts so don't worry about rebasing them for a little while. When I'm ready to merge this, I'll rebase the PR myself and make sure all the previous changes are applied in this PR as well. 🫡

lorisleiva:
@trevor-cortex

src/fragments/constantPage.ts
```diff
@@ -0,0 +1,39 @@
...
+): Fragment {
+    const typeManifest = visit(node.type, scope.typeManifestVisitor);
+    const rawValue = visit(node.value, scope.typeManifestVisitor).value;
+    const useBigInt = isNode(node.value, 'numberValueNode') && typeManifest.strictType.content === 'bigint';
+    const value = useBigInt ? fragment`${rawValue}n` : rawValue;
+    const valueType = isNode(node.value, 'stringValueNode')
+        ? fragment`string`
+        : useBigInt
+          ? fragment`bigint`
+          : typeManifest.strictType;
```
trevor-cortex: The `strictType.content === 'bigint'` check only catches direct number types. If the constant's type is a `definedTypeLinkNode` pointing at a `u64`/`i64` defined type, `strictType.content` is the link's type name (e.g. `BigNum`), so `useBigInt` is `false` and the value renders as a plain number literal:

```ts
export const X: BigNum = 42; // TS2322: number is not assignable to bigint
```

…which is a compile error in the generated client. Consider resolving the link through `linkables` to inspect the underlying number format before deciding on the `n` suffix, or falling back to `fragment`${rawValue}n`` whenever the resolved type lowers to `bigint`. Either way, a test with `constantNode('x', definedTypeLinkNode('bigNum'), numberValueNode(42))` where `bigNum` is a `u64` would lock this down.

src/visitors/getRenderMapVisitor.ts
```diff
@@ -141,6 +143,7 @@ export function getRenderMapVisitor(
                 },
 
                 visitProgram(node, { self }) {
+                    const constantsToExport = (node.constants ?? []).filter(isNotInternal);
```
trevor-cortex, line 146: Small semantics note: the docs describe `internalNodes` as "render nodes but omit them from barrel exports", but internal constants here are omitted from the rendered file entirely. I think full omission is actually the only sane option — the file is re-exported wholesale via `export *`, and an unexported `const` would trip `noUnusedLocals` — but it's a deliberate divergence worth a brief mention in the docs or a comment, so the next person doesn't "fix" it back.

trevor-cortex (commented):
This PR adds program constants generation: `program.constants` now renders into per-program files under a top-level `constants/` directory, with a barrel index and a root re-export. Constants are typed from their Codama type via the type manifest visitor, with special-casing for `bigint`-backed numbers (appending the `n` suffix) and string values (forcing `string` to avoid importing a declared type that doesn't match the literal). Sorting, docblocks, `internalNodes` filtering, import-extension handling, and the e2e anchor fixture are all covered.

The implementation follows the repo's patterns well — `nameApi.constant` instead of hardcoded naming, manifests instead of re-derived type logic, `getIndexPageFragment` reuse for the barrel, and a CLI-generated changeset with a `minor` bump that fits a new feature.

**Main thing to watch out for:** the `bigint` detection compares `strictType.content === 'bigint'`, which only catches direct number types. A constant whose type is a `definedTypeLinkNode` resolving to `u64`/`i64` will render `export const X: BigNum = 42;` — a TS compile error in the generated client since the alias resolves to `bigint` but the literal is a `number`. Details in the inline comment.

- Value rendering delegates entirely to the type manifest visitor, so non-scalar constant values (public keys, bytes, arrays, structs) will render whatever the visitor produces for them. None of those are covered by tests yet — worth a quick sanity check that at least `publicKeyValueNode` and `bytesValueNode` constants produce valid output, since Codama IDLs can express them.
- The e2e anchor fixture renders Anchor byte-array constants as their stringified form (e.g. `'[103, 117, ...]'`). That's faithful to how Anchor encodes constant values in the IDL, so it's correct for this PR, but consumers wanting actual seed bytes will need an upstream representation change — not something to fix here.
- The `constants/index.ts` barrel correctly stays absent when no program has exportable constants (`getIndexPageFragment` returns `undefined` for empty lists), and the root index only re-exports `./constants` when `programsWithConstantsToExport` is non-empty — the two are consistent.
- Per the repo's README skill, consider documenting the new `constants/` output directory in the package README as a follow-up, since this changes the generated tree structure users see.

trevor-cortex (approved):
This revision addresses both substantive points from my previous review:

- **Linked bigint types**: `resolvesToBigInt` in `constantPage.ts` now follows `definedTypeLinkNode` chains through `linkables` (with cycle protection via the `visited` set and correct path scoping when recursing into linked programs). The new test `it renders bigint literals for linked bigint types` locks down the previously broken case — `definedTypeLinkNode('bigNumber')` → `u64` now renders `export const MAX_AMOUNT: BigNumber = 42n;`. ✅
- **Non-scalar values**: new coverage for `publicKeyValueNode` (renders `Address` + `address(...)` with the right `@solana/kit` imports) and `bytesValueNode` (renders `ReadonlyUint8Array` + `new Uint8Array([...])`). ✅

Also a nice improvement beyond what I asked for: internal-node filtering is now applied at the constant level (`constantsToExport = (node.constants ?? []).filter(isNotInternal)`), and `programsWithConstantsToExport` uses `.some(isNotInternal)` so a program whose constants are *all* internal correctly produces no constants file and no barrel entry. The hoisted `isNotInternal` keeps `visitProgram` and `visitRoot` consistent.

One non-blocking edge case for a potential follow-up (not caused by this PR): the bigint check compares `strictType.content === 'bigint'`, so branded bigint aliases produced by wrapper types (e.g. an amount type rendering as `Lamports`) won't match, and even with an `n` suffix a bare literal wouldn't satisfy a branded type without its helper (`lamports(...)`). Constants with amount-style types are probably rare-to-nonexistent in real IDLs, and fixing it properly belongs in value rendering rather than here — just flagging it so it's a known limitation.

Remaining follow-up from last time still stands: consider mentioning the new `constants/` output directory in the package README. Not blocking.

LGTM — nice, tight revision.

lorisleiva (approved):
Thank you!

PR #176, opened 2026-08-01
Group type-only imports and fix missing type keyword on PDA seeds imports

macalinao:
Two small fixes that make generated clients cleaner under `verbatimModuleSyntax`, which every e2e fixture tsconfig already enables.

## Group type-only imports

When every identifier a generated file imports from a given module is type-only, the import is now rendered as a single `import type { ... }` statement rather than inlining a `type` marker on each specifier. The two forms are equivalent to the type checker, but they differ in what TypeScript emits: under `verbatimModuleSyntax` the inline form erases the specifiers and leaves an empty `import {} from '...'` side-effect statement behind in the emitted JavaScript, while the grouped form is erased entirely. Modules that mix value and type imports keep the existing inline form, since only a fully type-only group can be hoisted to `import type`.

## Fix the PDA seeds type import

`getAccountPdaHelpersFragment` imported the PDA seeds type (e.g. `BarSeeds`) without a `type` keyword, even though it is only ever referenced in type position on the `seeds` parameter. Any generated client with an account linked to a PDA that has variable seeds therefore failed to compile with TS1484 under `verbatimModuleSyntax`. The identifier is now imported as `type BarSeeds`; `mergeImportMaps` already lets a value import win over a type import, so this stays correct if the same name is also needed as a value.

## Notes for consumers

Regenerating a client will produce cosmetic `import { type A }` → `import type { A }` diffs. No runtime behavior changes.

The checked-in e2e fixtures were regenerated as part of this change to keep CI's clean-working-tree invariant satisfied; the only resulting diff is a single import statement in the dummy client.

## Related work

Part of #180 — running generated clients natively on Node.js 24+ TypeScript (type stripping), with no separate compilation step:

- #176: `import type` grouping and the missing `type` keyword fix (`verbatimModuleSyntax` polish)
- #177: `importExtension` option — explicit `.js`/`.ts` extensions on relative imports
- #178: `erasableSyntax` option — erasable replacements for generated `enum` declarations
- #179: `modern` e2e fixture generated with both options (draft, stacked on #177 and #178)

### DIFF
--- a/src/utils/importMap.ts
+++ b/src/utils/importMap.ts
@@ -147,11 +147,17 @@ export function importMapToString(
             return a.localeCompare(b);
         })
         .map(([module, imports]) => {
-            const innerImports = [...imports.values()]
-                .map(importInfoToString)
+            const importInfos = [...imports.values()];
+            // When every import of a module is type-only, we can use a single
+            // `import type` statement. This ensures the statement is fully erased
+            // from the emitted JavaScript under `verbatimModuleSyntax`, instead of
+            // leaving an empty `import {} from '...'` side-effect statement behind.
+            const isTypeOnlyModule = importInfos.length > 0 && importInfos.every(({ isType }) => isType);
+            const innerImports = importInfos
+                .map(importInfo => importInfoToString(importInfo, isTypeOnlyModule))
                 .sort((a, b) => a.localeCompare(b))
                 .join(', ');
-            return `import { ${innerImports} } from '${module}';`;
+            return `import ${isTypeOnlyModule ? 'type ' : ''}{ ${innerImports} } from '${module}';`;
         })
         .join('\n');
 }
@@ -198,7 +204,10 @@ function resolveImportMapModules(
     );
 }
 
-function importInfoToString({ importedIdentifier, isType, usedIdentifier }: ImportInfo): string {
+function importInfoToString(
+    { importedIdentifier, isType, usedIdentifier }: ImportInfo,
+    omitTypeKeyword: boolean = false,
+): string {
     const alias = importedIdentifier !== usedIdentifier ? ` as ${usedIdentifier}` : '';
-    return `${isType ? 'type ' : ''}${importedIdentifier}${alias}`;
+    return `${isType && !omitTypeKeyword ? 'type ' : ''}${importedIdentifier}${alias}`;
 }
--- a/test/_setup.ts
+++ b/test/_setup.ts
@@ -162,13 +162,13 @@ export function fragmentDoesNotContainImports(
 }
 
 export async function codeContainsImports(actual: string, expectedImports: Record<string, string[]>): Promise<void> {
-    const normalizedActual = await inlineCode(actual);
+    const actualImports = await getImportPairs(actual);
     const importPairs = Object.entries(expectedImports).flatMap(([key, value]) => {
         return value.map(v => [key, v] as const);
     });
 
-    importPairs.forEach(([importFrom, importValue]) => {
-        expect(normalizedActual).toMatch(new RegExp(`import{[^}]*\\b${importValue}\\b[^}]*}from'${importFrom}'`));
+    importPairs.forEach(importPair => {
+        expect(actualImports).toContainEqual(importPair);
     });
 }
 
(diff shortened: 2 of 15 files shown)


lorisleiva:
@trevor-cortex

src/utils/importMap.ts
```diff
@@ -147,11 +147,17 @@ export function importMapToString(
...
         })
         .map(([module, imports]) => {
-            const innerImports = [...imports.values()]
-                .map(importInfoToString)
+            const importInfos = [...imports.values()];
+            // When every import of a module is type-only, we can use a single
+            // `import type` statement. This ensures the statement is fully erased
+            // from the emitted JavaScript under `verbatimModuleSyntax`, instead of
+            // leaving an empty `import {} from '...'` side-effect statement behind.
+            const isTypeOnlyModule = importInfos.every(({ isType }) => isType);
```
trevor-cortex: Nit: `Array.prototype.every` returns `true` for an empty array, and an empty module entry is representable (e.g. `addToImportMap(map, 'm', [])` sets the module key with an empty specifier map via `mergeImportMaps`). Such an entry would previously render as `import { } from 'm'` — a side-effect import — and would now render as `import type { } from 'm'`, which gets erased entirely. No current caller passes an empty specifier list, so this is purely theoretical, but if you want to be defensive: `const isTypeOnlyModule = importInfos.length > 0 && importInfos.every(({ isType }) => isType);` preserves the old behavior for that edge.

test/_setup.ts
```diff
@@ -168,7 +168,9 @@ export async function codeContainsImports(actual: string, expectedImports: Recor
     });
 
     importPairs.forEach(([importFrom, importValue]) => {
-        expect(normalizedActual).toMatch(new RegExp(`import{[^}]*\\b${importValue}\\b[^}]*}from'${importFrom}'`));
+        expect(normalizedActual).toMatch(
+            new RegExp(`import(?: type)?\\{[^}]*\\b${importValue}\\b[^}]*\\}from'${importFrom}'`),
```
trevor-cortex: Observation, not a request: making the marker optional with `(?: type)?` means these helpers can no longer distinguish the two forms — an assertion for a *value* import would still pass if the identifier were accidentally rendered inside an `import type` statement (which would break at runtime). That trade-off seems fine given the form is pinned down precisely in `test/utils/importMap.test.ts` and the e2e fixtures compile the output, but worth keeping in mind if a test ever needs to assert value-ness specifically.

trevor-cortex (approved):
This PR makes two related `verbatimModuleSyntax` fixes: (1) `importMapToString` now hoists the `type` keyword to the import statement level (`import type { ... }`) when every specifier for a module is type-only, so the statement is fully erased from emitted JS instead of leaving an empty side-effect `import {} from '...'` behind; (2) `getAccountPdaHelpersFragment` now imports the PDA seeds type with the `type` marker, fixing a TS1484 compile error in generated clients that have accounts linked to variable-seed PDAs.

The logic is sound:

- The grouping only applies when *all* specifiers are type-only, so mixed modules correctly keep inline markers — verified by the new mixed-module tests.
- The seeds-type fix is safe against future collisions because `mergeImportMaps` already lets a value import win over a type import for the same identifier.
- Sort order of specifiers is unaffected: for type-only modules every string previously shared the `type ` prefix, so stripping it doesn't reorder anything; mixed modules are untouched.
- The changeset is a `patch` for a fix, consistent with the repo's changeset conventions, and the dummy e2e fixture was regenerated to satisfy CI's clean-working-tree invariant.

- Worth double-checking that *all* checked-in e2e fixtures (`anchor`, `system`, `memo`, `token`, not just `dummy`) were regenerated — any generated file whose entire import from a module is type-only will now diff, and a missed one would trip the clean-tree check in CI. The PR states only the dummy client changed, which is plausible (most instruction files mix value imports like `getStructEncoder` with types in the same statement), but it's cheap to verify locally.
- The relaxed regex in `test/_setup.ts` means the import assertion helpers now accept either form (see inline comment) — type-only rendering correctness is instead pinned down by the `importMap.test.ts` unit tests and e2e compilation.

Two minor inline observations below, neither blocking.

trevor-cortex (approved):
Re-reviewed after the latest changes — still approved. Both points from my previous review have been addressed:

1. **Empty specifier list guard**: `isTypeOnlyModule` now requires `importInfos.length > 0`, and the new `it preserves empty import statements` test pins the behavior so a hypothetical empty `import { } from 'm'` side-effect statement can never be silently erased into `import type { }`.
2. **Test helpers can assert value-ness again**: `codeContainsImports` / `codeDoesNotContainImports` no longer rely on a permissive regex. The new `getImportPairs` helper parses each import statement into `(module, specifier)` pairs and canonicalizes both `import type { A }` and `import { type A }` to `type A`, while a plain `A` expectation now only matches a genuine value import. This is strictly better than what I'd flagged — the helpers are now *exact* on type-vs-value, which is why the follow-up diffs across `programsPage.test.ts`, `definedType.test.ts`, `zeroableOption.test.ts`, etc. add the `type ` prefixes to expectations that were previously under-specified. The new `test/utils/setup.test.ts` meta-tests lock the helper semantics in, which is a nice touch given how much of the suite leans on them.

Remaining logic is unchanged from my prior pass: the grouping only fires when every specifier for a module is type-only, mixed modules keep inline markers, sort order is unaffected, and the PDA seeds `type` fix remains safe against future value-import collisions via `mergeImportMaps` precedence.

- `getImportPairs` only recognizes named-import statements (`import [type] { ... } from '...'`) in `inlineCode`-normalized form. That covers everything the renderer emits today, but if generated code ever grows default or namespace imports, the helpers will silently not see them — worth remembering if import-assertion tests ever pass suspiciously.
- The earlier point about verifying the other e2e fixtures (`anchor`, `system`, `memo`, `token`) for missed type-only-import diffs still applies; CI's clean-working-tree check is the backstop.

lorisleiva (approved):
Thank you!

PR #177, opened 2026-08-01
Add importExtension option for explicit extensions on relative imports

macalinao:
This PR adds an opt-in `importExtension?: "js" | "ts"` render option for generated relative imports and re-exports. With `"js"`, generated clients work with Node ESM and TypeScript `moduleResolution: "NodeNext"`; with `"ts"`, specifiers point directly at TypeScript source files for runtimes that execute them. Leaving the option unset preserves the existing extensionless output.

## How it works

The render options build a shared `getImportPath` helper on `RenderScope`. Internal imports and barrel exports use that helper consistently, whilst external packages and user-provided dependency or link paths remain unchanged. Renderer-owned directory aliases resolve to `index.<ext>` and explicit file extensions are preserved. If a renderer-owned alias is overridden to a file, that override should include its extension so it is not treated as a directory.

```ts
// With `importExtension: "js"`.
import type { MyType } from "../types/index.js";
export * from "./myAccount.js";
```

## Tests

Unit tests cover JavaScript and TypeScript extensions, generated files and barrels, sibling type imports, instructions, external packages, user overrides and the unchanged default behaviour.

A minimal NodeNext e2e fixture provides real compiler and runtime coverage without requiring a validator. It generates accounts, errors, PDAs, programs and types, compiles the client as ESM using `module` and `moduleResolution` set to `NodeNext`, then verifies barrel exports, a codec round trip and PDA derivation through the compiled JavaScript.


lorisleiva:
@trevor-cortex

trevor-cortex (commented):
## Summary

Adds an opt-in `importExtension?: 'js' | 'ts'` render option that appends explicit extensions to relative specifiers in generated code — `./myAccount.<ext>` for files, `./folder/index.<ext>` for the renderer's own directory aliases. The rewrite is applied at the symbolic-module-key level via a `getImportPath` function threaded through `RenderScope`, so user-supplied `dependencyMap`/`linkOverrides` paths and non-relative specifiers are emitted verbatim. A new `nodenext` e2e project compiles a generated `importExtension: 'js'` client under `module`/`moduleResolution: NodeNext` and runs offline AVA tests through Node's own resolver.

The core design is sound: rewriting on the alias keys (not resolved paths) is the right pivot — it's what keeps the per-file `{ generatedTypes: '.' }` override resolving to `./index.js` while leaving genuinely user-supplied paths alone, and the `RECOGNIZED_EXTENSION_REGEX` escape hatch gives users a clean way to point an alias at a file. Default behavior is unchanged (default `getImportPathFactory()` is the identity for extensionless output), and `getExternalDependencies` continuing to resolve without extensions is harmless since relative modules are filtered out there anyway. Unit coverage in `importMap.test.ts` and `importExtension.test.ts` is thorough and hits the right edges (overrides with extensions, npm-package overrides, unowned relative modules, untouched default).

## Things to resolve

1. **`test/e2e/nodenext/.eslintrc.cjs` points at a `tsconfig.json` that doesn't exist in that package** — the build uses `../tsconfig.nodenext.json` and no local tsconfig is checked in, so the package-local `pnpm lint` script will fail to resolve the parser project. See inline comment.
2. **The e2e fixture doesn't match the PR description.** The description says the fixture is anchor-based and "produces all six generated folders" in twenty files, but the checked-in `idl.json` is a slim counter IDL with `"instructions": []` — no `instructions/` folder is generated, so `generatedInstructions` (the richest import surface: shared helpers, program constants, type links) is never exercised through real NodeNext resolution. Either add a small instruction to the IDL or update the description. See inline comment.
3. **Scope question:** the `src/fragments/accountPdaHelpers.ts` hunk (adding the `type` keyword to the PDA seeds import) and the accompanying `accountsPage.test.ts` test read like they belong to #176 ("the missing `type` keyword fix") rather than this PR. If #177 is stacked on #176 and this is just the diff being computed against `main`, all good — otherwise consider moving those hunks to keep this PR single-purpose.

## Notes for maintainer sign-off

On the edge case called out in the description — a renderer-owned directory alias overridden to an extensionless *file* path (`{ generatedTypes: '../myTypes' }`): I'd keep the current behavior. The alias contract is "directory with an index file", the mitigation (include the extension) is documented, and narrowing the rule so overrides are never rewritten would break the `{ generatedTypes: '.' }` per-file case that keeps sibling type imports working. Current trade-off looks right.

Smaller notes, no action needed:

- `test/e2e/tsconfig.nodenext.json` living in the e2e root with hard-coded `nodenext/` paths is a little unusual next to the sibling projects' local tsconfigs, but it works; a local `tsconfig.json` extending it would also resolve point 1 above.
- `getDefaultInternalModuleMap` is now rebuilt per page render instead of being a module constant — negligible cost, and it keeps the factory injection clean.
- The `dependencyMap` → `pageDependencyMap` shadowing rename in `asPage` is a nice cleanup.
- Changeset is a `minor` for a new opt-in feature, consistent with repo precedent.

trevor-cortex (approved):
## Re-review

All three points from my previous review have been addressed:

1. **ESLint config fixed** — `test/e2e/nodenext/.eslintrc.cjs` now points at `../tsconfig.nodenext.json` (which resolves correctly given `tsconfigRootDir: __dirname`), and the root `eslint.config.mjs` gained a matching override for `test/e2e/nodenext/**/*.ts` with `projectService: false`. Both paths verified against the checked-in `tsconfig.nodenext.json`.
2. **Fixture/description mismatch resolved** — the PR description now accurately describes a minimal fixture (accounts, errors, PDAs, programs, types; no instructions), and `test/importExtension.test.ts` gained an `increment` instruction in its test node, asserting `instructions/increment.ts` imports `../programs/index.js` and that the root index re-exports `./instructions/index.js`. Specifier-level coverage of the instructions surface now exists at the unit level.
3. **Scope cleaned up** — the `accountPdaHelpers.ts` `type`-keyword hunk and its `accountsPage.test.ts` companion are no longer in this diff (34 → 32 files), keeping the PR single-purpose.

## Verified on this pass

- The build wiring holds together: `tsc -p ../tsconfig.nodenext.json` with `rootDir: ./nodenext` / `outDir: ./nodenext/dist` produces `dist/src` and `dist/test`, which matches AVA's `rewritePaths: { 'test/': 'dist/test/' }` and the compiled test's `../src/index.js` specifier.
- `RECOGNIZED_EXTENSION_REGEX` covers `.js`/`.ts`/`.jsx`/`.tsx`/`.mjs`/`.cjs`/`.mts`/`.cts`/`.json` — the sensible set for the "override already carries an extension" escape hatch.
- Default behavior remains the identity path (`getImportPathFactory()` with no extension), confirmed by the untouched-default unit test.

## Notes for maintainers

- The `nodenext` e2e fixture intentionally omits instructions, so the `generatedInstructions` alias is exercised through unit tests only, not through real NodeNext resolution. That's a reasonable trade-off for a minimal fixture, and #179's `modern` fixture can pick up the slack if fuller e2e coverage is wanted.
- My recommendation on the directory-alias-overridden-to-a-file edge case stands: keep the current behavior. The alias contract is "directory with an index file", the extension-in-override mitigation is documented, and narrowing would break the `{ generatedTypes: '.' }` per-file case.
- Trivial, non-blocking: `getImportPathFactory()` is defaulted in both `importMapToString` and `resolveImportMapModules` — harmless redundancy since the former always passes it through.

lorisleiva (approved):
Thank you!

PR #178, opened 2026-08-01
Add erasableSyntax option to generate erasable enum alternatives

macalinao:
## Motivation

This PR adds an opt-in `erasableSyntax` option for generated clients consumed by type-stripping toolchains. TypeScript's `erasableSyntaxOnly` flag and Node.js type stripping reject syntax that carries runtime behaviour, and generated clients currently use two affected constructs: `enum` declarations and angle-bracket type assertions.

Together with `importExtension: 'ts'` from #177, this enables generated clients to run directly from their TypeScript sources on Node.js 24+ without `--experimental-transform-types` or a separate compilation step.

## What changes

When enabled, numeric enums become const objects paired with union types of the same name:

```ts
export const Key = {
    0: 'Uninitialized',
    1: 'Asset',
    Uninitialized: 0,
    Asset: 1,
} as const;

export type Key = (typeof Key)[Exclude<keyof typeof Key, number>];
```

The object preserves the runtime shape of a numeric enum, including its reverse mappings. This is required because `@solana/codecs` inspects that shape when building enum codecs. Keeping the numeric properties in the exported object type also means consumers can use forward and reverse lookups in TypeScript, whilst `Exclude<keyof typeof Key, number>` limits the enum's value type to its forward mapping values.

The existing codec calls remain unchanged:

```ts
getEnumEncoder(Key);
getEnumDecoder(Key);
```

The option covers all three places enums are emitted: scalar defined types, program account enums and program instruction enums. Enum members referenced from a type position use a `typeof` query because a const object only exists in the value space.

The program plugin object also switches from an angle-bracket assertion to an equivalent `as` assertion when the option is enabled, since angle-bracket assertions are rejected by `erasableSyntaxOnly`.

## Compatibility

Everything is gated behind the new option, which defaults to `false`, so existing generated output remains unchanged. The option is resolved into a required boolean on `RenderScope`, consistent with the renderer's other resolved options.

## Verification

- The full unit suite passes across the Node, browser and React Native projects: 585 tests.
- The TypeScript typecheck, build, ESLint and Prettier checks pass.
- The generated enum shape was checked against the installed Solana codec types and successfully round-tripped `Key.Asset` to byte `1` and back, with its reverse lookup preserved.

## Notes for review

- Variant values are positional indices used by the codecs as discriminators. Explicit `discriminator` values on `enumEmptyVariantTypeNode` remain ignored, matching the existing native enum output.

## Related work

Part of #180, which tracks running generated clients natively on Node.js 24+ TypeScript:

- #176: type-only import grouping and the missing `type` keyword fix
- #177: explicit `.js` and `.ts` extensions for relative imports
- #178: erasable alternatives for generated enums and assertions
- #179: a modern E2E fixture generated with both options


lorisleiva:
@trevor-cortex

src/fragments/enumDeclaration.ts
```diff
@@ -0,0 +1,59 @@
...
+        name: string;
+    },
+): Fragment {
+    const { name, body, docblock = fragment``, erasableSyntax } = scope;
+
+    if (!erasableSyntax) {
+        return fragment`${docblock}export enum ${name} ${body}`;
+    }
+
+    return fragment`${docblock}export const ${name} = ${body} as const;\n\nexport type ${name} = (typeof ${name})[Exclude<keyof typeof ${name}, number>];`;
```
trevor-cortex: Nit: the docblock only attaches to the `const` declaration here, so consumers referencing `Key` in a type position won't see the docs on hover. Consider emitting the docblock before both declarations (or at least before the type alias too), since the two share a name and either could be the one a reader lands on.

src/visitors/getTypeManifestVisitor.ts
```diff
@@ -296,12 +306,13 @@ export function getTypeManifestVisitor(input: {
                             );
                         }
                         const variantNames = (enumType.variants ?? []).map(({ name }) => nameApi.enumVariant(name));
+                        const body = erasableSyntax ? getErasableEnumBody(variantNames) : variantNames.join(', ');
```
trevor-cortex: Nit: this ternary duplicates the exact body-selection logic in `getEnumDeclarationFragment` (`src/fragments/enumDeclaration.ts` L18). Consider folding the `erasableSyntax` switch into a single shared helper — e.g. `getEnumBody(variantNames, erasableSyntax)` next to `getErasableEnumBody` — so the two call sites can't drift apart.

trevor-cortex (approved):
This PR adds an opt-in `erasableSyntax` option (default `false`) that makes generated clients compatible with TypeScript's `erasableSyntaxOnly` and Node.js type stripping. When enabled, the three enum emission sites (scalar defined types, program account enums, program instruction enums) render a `const` object + same-named union type instead of a native `enum`, and the program plugin object switches from an angle-bracket assertion to an `as` assertion. The option is resolved into a required boolean on `RenderScope` and threaded through the affected fragments.

The approach is sound and I verified the tricky bits:

- The const object body (reverse mappings followed by forward entries) matches the runtime shape of a compiled numeric enum, which is what `@solana/codecs` inspects in `getEnumEncoder`/`getEnumDecoder` — and the codec call sites are unchanged.
- `Exclude<keyof typeof X, number>` correctly strips the numeric reverse-mapping keys, so the exported type is the same `0 | 1 | …` value union a modern (TS 5+) numeric enum produces. No meaningful assignability change for consumers.
- I traced the other enum reference sites: `visitEnumValue` emits members in value space only (works against the const), the account identifier function's return type annotation resolves via the same-named type alias, and the parsed instruction union — the one place an enum *member* appears in a type position — correctly gets the `typeof` query. Declaration merging of `export const X` / `export type X` also survives the `export *` barrels.
- Gating is clean: with the option off, output is byte-for-byte what it was before, and the default-path tests pin that down.

- The main thing worth double-checking is completeness: whether any *other* non-erasable construct (angle-bracket assertion, namespace, parameter property) still exists elsewhere in the generated output. The author reports typechecking generated output and #179 adds an E2E fixture with the option enabled, which should catch stragglers — worth confirming that fixture compiles under `erasableSyntaxOnly` in CI.
- `getErasableEnumBody` single-quotes variant names without escaping. Variant names come from `nameApi.enumVariant` (identifier-shaped, since they must also be valid native enum members), so this isn't exploitable today — just something to keep in mind if the name API ever loosens.

Two minor nits inline; nothing blocking.

trevor-cortex (approved):
Re-reviewed after the follow-up commit — both nits from my previous review are addressed, and I have no new concerns. Still approved.

**What changed since my last pass:**

- The docblock is now emitted before both the `const` declaration and the `type` alias in `getEnumDeclarationFromBodyFragment`, so hover documentation is preserved whether the enum name is used in a value or a type position. The scalar enum test pins this by asserting the docblock appears on both declarations.
- The duplicated `erasableSyntax ? … : variantNames.join(', ')` logic is now a single shared `getEnumBody` helper in `src/utils/erasableSyntax.ts`, used by both `getTypeManifestVisitor` and `getEnumDeclarationFragment`. Its docblock also documents why the reverse-mapping entries precede the forward entries (matching the compiled numeric enum shape that `@solana/codecs` introspects), which is a nice touch.

Everything else is unchanged from the version I already verified: the const object's runtime shape, the `Exclude<keyof typeof X, number>` value union, the `typeof` query for the one type-position member reference in the parsed instruction union, the `as` assertion for the plugin object, and byte-for-byte identical output when the option is off.

**Still worth confirming downstream** (carried over from my previous review, not a blocker here): that #179's E2E fixture actually compiles under `erasableSyntaxOnly` in CI, as the end-to-end guard against any remaining non-erasable construct in generated output.

lorisleiva (approved):
Thank you!

PR #179, opened 2026-08-01
Consolidate Node ESM compatibility tests

macalinao:
This PR adds focused Node ESM compatibility coverage for explicit TypeScript import extensions and erasable syntax. A small generated client is compiled under strict NodeNext settings, executed from emitted JavaScript, and executed directly from TypeScript under Node type stripping. It also fixes enum decoder inference uncovered by the consumer-level typecheck.

The new `test/e2e/node-esm` fixture replaces the existing `nodenext` fixture and consolidates the smaller compatibility fixture developed in this PR. It uses one scalar enum and one instruction to cover typed enum reverse lookups, codec round-trips, instruction identification and parsing, and the program plugin. Unit tests continue to cover the renderer's explicit `.js` import output, whilst this fixture exercises the more demanding `.ts` source-to-`.js` emission path end to end.

No additional changeset is required because the existing pending minor changeset for `erasableSyntax` already covers the corrected generated decoder output.


lorisleiva:
@trevor-cortex

src/visitors/getTypeManifestVisitor.ts
```diff
@@ -307,8 +307,11 @@ export function getTypeManifestVisitor(input: {
                         }
                         const variantNames = (enumType.variants ?? []).map(({ name }) => nameApi.enumVariant(name));
                         const body = getEnumBody(variantNames, erasableSyntax);
+                        const decoderConstructor = erasableSyntax
+                            ? fragment`${currentParentName.strict} as Omit<typeof ${currentParentName.strict}, number>`
+                            : fragment`${currentParentName.strict}`;
```
trevor-cortex, line 314: Nit: a one-line comment here would help future readers — e.g. that the erasable const object carries numeric reverse mappings the codec needs at runtime, but the *type* must exclude the numeric keys so the decoder infers the variant value union rather than `Key | 'Uninitialized' | 'Asset'`. Also worth noting why the encoder on the line below doesn't need the same narrowing (encoder inputs are contravariant, so the wider inferred type is still assignable to `FixedSizeEncoder<KeyArgs>`). The unit test explains this, but the visitor is where people will read it first.

test/e2e/node-esm/README.md
```diff
@@ -0,0 +1,33 @@
+# `node-esm` e2e fixture
+
+Proves that a generated client can be executed directly by Node.js 24+ through
```
trevor-cortex: Minor inconsistency: this says Node.js **24+**, but `run-type-stripping.cjs` (correctly) notes type stripping is available unflagged from Node **22.18 / 23.6** onwards, and the gate is `process.features.typescript` rather than a version check. Suggest aligning the two — e.g. "Node.js 22.18+ (via type stripping)" or simply "Node versions that support type stripping".

trevor-cortex (commented):
This PR consolidates the `nodenext` E2E fixture into a new `node-esm` fixture that exercises the two most demanding renderer options together (`importExtension: 'ts'` + `erasableSyntax: true`) under a strict NodeNext/`erasableSyntaxOnly` tsconfig, runs the compiled output, and additionally runs the TypeScript source directly via Node type stripping. It also fixes the enum decoder inference for erasable enums by narrowing the const object with `Omit<typeof X, number>` so the decoded type is the variant value union rather than including the reverse-mapping strings.

The core fix looks correct to me:

- `as` assertions are erasable, so the narrowing survives `erasableSyntaxOnly`, and the runtime object keeps the numeric reverse mappings `getEnumDecoder` needs.
- Leaving the **encoder** un-narrowed is fine — encoders are contravariant in their input, so an encoder inferred from the wider object is assignable to `FixedSizeEncoder<KeyArgs>`. The asymmetry is deliberate and the updated unit test documents it.
- The fixture wiring is consistent end to end: `generate.cjs` options, `test.sh`, the pnpm workspace importer rename, and the self-contained `tsconfig.json` (the `.eslintrc.cjs` now pointing at a local `tsconfig.json` is a nice decoupling).

Things to address / verify:

1. **Orphaned shared tsconfig.** `test/e2e/tsconfig.nodenext.json` still exists at this PR's head and references the now-deleted `nodenext` project (`rootDir: ./nodenext`, `include: ["nodenext/src", "nodenext/test"]`). Nothing references it anymore — the new fixture has its own local `tsconfig.json`. It should be deleted in this PR to avoid leaving dead config behind.
2. **PR description vs. diff.** The description says this fixture replaces the separate `nodenext` and `modern` fixtures, but only `nodenext` is deleted here and I can't find a `modern` fixture on `main` or at the PR head. If `modern` was handled in another PR in the stack, fine — otherwise the description is slightly off.
3. **Changeset.** The claim that the pending `erasableSyntax` minor changeset covers the corrected decoder output is reasonable (the feature hasn't shipped, so the fix folds into it), but a subsequent reviewer may want to confirm that changeset is still pending in `.changeset/`.

For subsequent reviewers: the generated files under `test/e2e/node-esm/src/generated/` are checked-in renderer output, and CI enforces a clean working tree after tests — so the main thing to verify is that regenerating via `test/e2e/generate.cjs node-esm` reproduces them byte-for-byte with the new visitor change.

trevor-cortex (approved):
Re-reviewed after the latest revision — all three points from my previous review are addressed:

1. **Orphaned shared tsconfig**: resolved. `test/e2e/tsconfig.nodenext.json` no longer exists at head — it's now repurposed as the fixture-local `test/e2e/node-esm/tsconfig.json` (tracked as a rename), with `rootDir`/`outDir`/`include` updated accordingly. No dead config left behind.
2. **Description mismatch**: resolved. The description no longer claims a separate `modern` fixture is being replaced; it now accurately says the PR consolidates `nodenext` with the compatibility fixture developed here.
3. **Decoder/encoder asymmetry comment**: added in `getTypeManifestVisitor.ts` — the new comment explains exactly why only the decoder is narrowed, which is what a future reader will need.

The README also no longer overstates the Node version requirement; `run-type-stripping.cjs` gates on `process.features.typescript` with the correct 22.18/23.6 framing.

The substance of the change remains as previously assessed: the `Omit<typeof X, number>` assertion is erasable, keeps the runtime reverse mappings `getEnumDecoder` needs, and narrows the decoded type to the variant value union, while the encoder correctly stays un-narrowed (contravariant input). The fixture wiring is consistent end to end (`generate.cjs`, `test.sh`, pnpm workspace importer, local eslint/tsconfig), and the updated `erasableSyntax.test.ts` documents the new output.

One remaining note for the merger (no action needed in the diff): I couldn't enumerate `.changeset/` contents to confirm the pending `erasableSyntax` minor changeset is still present — worth a quick glance that it's there before merge, since this PR relies on it covering the corrected decoder output.

LGTM.

lorisleiva (approved):
Thank you!

PR #188, opened 2026-08-18
fix: forward program address to linked PDA default values

amilz:
## Problem

Instruction accounts defaulting to a **linked** PDA were rendered as `await findFooPda(seeds)` — no config. Since `findFooPda` bakes in `pdaNode.programId ?? programNode.publicKey`, passing `{ programAddress }` to the instruction builder produced an instruction aimed at the override with PDAs derived from the original address.

Visible in the SPL clients today: `getCreateAssociatedTokenInstructionAsync(input, { programAddress })` derives the ATA under `ATokenGPv…` regardless.

Ref: https://github.com/solana-program/token/blob/f5285693a93135a144e24859c84d26ac20037a3a/clients/js/src/generated/instructions/createAssociatedToken.ts#L160-L173

## Fix

Forward `{ programAddress }` to the find function when the linked PDA is derived from the instruction's own program. PDAs owned by another program, or pinned via an explicit `pdaNode.programId`, keep their own address.

This removes an inconsistency rather than adding behavior — the inlined-PDA branch already defaults to `programAddress`, and `fetch*FromSeeds` already forwards it.

## Notes

- Changes generated output for any client with a same-program linked PDA default; regenerate after upgrading.
- `linkOverrides.pdas` finders must now accept a trailing `config: { programAddress?: Address }`.
- Regenerated `test/e2e/anchor` + `test/e2e/token`.
- Pre-existing on `main`, untouched here: 5 eslint errors in `test/e2e/dummy/test/instructions.test.ts`.

### DIFF
--- a/.changeset/small-emus-happen.md
+++ b/.changeset/small-emus-happen.md
@@ -1,0 +1,11 @@
+---
+'@codama/renderers-js': minor
+---
+
+Forward the program address to linked PDA default values in async instruction builders
+
+Instruction accounts defaulting to a linked PDA were derived by calling the generated `find*Pda()` function with no config, so the derivation always used the address baked into that function and silently ignored the `programAddress` passed to the instruction builder. Programs deployed at different addresses per cluster produced instructions aimed at the overridden program with PDAs derived from the original one.
+
+Async builders now pass `{ programAddress }` through to the find function when the PDA is derived from the instruction's own program, matching what the generated `fetch*FromSeeds` account helpers already did. PDAs owned by another program, or pinned to an explicit program ID, keep their own address.
+
+This changes generated output for any client with a same-program linked PDA default, so regenerate after upgrading. PDA finders supplied through `linkOverrides.pdas` are called with the same signature as generated ones, so they must accept a trailing `config: { programAddress?: Address }` argument.
--- a/src/fragments/instructionInputDefault.ts
+++ b/src/fragments/instructionInputDefault.ts
@@ -1,7 +1,22 @@
 /* eslint-disable no-case-declarations */
-import { camelCase, InstructionInputValueNode, isNode, OptionalAccountStrategy } from '@codama/nodes';
+import {
+    camelCase,
+    InstructionInputValueNode,
+    InstructionNode,
+    isNode,
+    isNodeFilter,
+    OptionalAccountStrategy,
+    PdaNode,
+} from '@codama/nodes';
 import { mapFragmentContent, setFragmentContent } from '@codama/renderers-core';
-import { pipe, ResolvedInstructionInput, visit } from '@codama/visitors-core';
+import {
+    findProgramNodeFromPath,
+    getLastNodeFromPath,
+    NodePath,
+    pipe,
+    ResolvedInstructionInput,
+    visit,
+} from '@codama/visitors-core';
 
 import {
     addFragmentFeatures,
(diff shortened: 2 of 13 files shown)


lorisleiva:
@trevor-cortex

.changeset/small-emus-happen.md
```diff
@@ -0,0 +1,11 @@
+---
+'@codama/renderers-js': minor
```
trevor-cortex, line 2: Question on the bump level: the last paragraph documents that `linkOverrides.pdas` finders must now accept a trailing `config` argument — that's a compile-time break for TypeScript consumers whose override finders declare a narrower signature (generated call sites will pass two args). Is `minor` consistent with how this repo has treated changes to the `linkOverrides` contract before? Not asserting it should be `major` — just worth a quick check against precedent in past changesets before merging.

src/fragments/instructionInputDefault.ts
```diff
@@ -156,11 +182,13 @@ export function getInstructionInputDefaultFragment(
                 mergeFragments(pdaSeeds, renders => renders.join(', ')),
                 f => mapFragmentContent(f, c => `{ ${c} }`),
             );
-            if (pdaSeeds.length > 0) {
+            if (pdaSeeds.length > 0 || hasVariablePdaSeeds(pdaPath)) {
                 pdaArgs.push(pdaSeedsFragment);
             }
```
trevor-cortex, line 187: Good catch on the positional guard — verified against `pdaFunction.ts` that the seeds argument only exists when the PDA has variable seeds, so this keeps the config in the right slot. Bonus: this also fixes a pre-existing misposition in the `pdaProgramValue` branch below (explicit `programId` override with no seed values would previously have put the config where seeds belong).

One observation, no action needed: when variable seeds exist but no seed values are provided (and nothing fills them upstream), the rendered `findCounterPda({}, { programAddress })` won't satisfy the required seeds type — but the old `findCounterPda()` didn't type-check in that case either, so this surfaces the same misconfiguration at compile time, just with correct argument positions.

trevor-cortex (approved):
## Summary

This PR fixes async instruction builders silently ignoring a `programAddress` override when resolving linked-PDA account defaults. Previously the generated code called `await findFooPda(seeds)` with no config, so the derivation always used the address baked into the finder (`pdaNode.programId ?? programNode.publicKey`) even when the instruction was targeting an overridden program address. The fix forwards `{ programAddress }` when — and only when — the linked PDA belongs to the instruction's own program and isn't pinned via an explicit `pdaNode.programId`, resolved through `linkables.getPath(...)`.

I verified the assumptions against `src/fragments/pdaFunction.ts`: generated finders take `(seeds?, config = {})` with the seeds argument present only when there are variable seeds, so the new `hasVariablePdaSeeds` guard correctly keeps the config argument in the right position. The gating in `isPdaDerivedFromInstructionProgram` is also correct on both exclusion axes: cross-program PDAs keep their own baked-in address, and explicitly pinned PDAs must not be overridden (forwarding would clobber the pin, since the finder's `programAddress` config wins over the default). This also brings the linked-PDA branch in line with the inlined-PDA branch (which already defaults to `programAddress`) and with `fetch*FromSeeds`.

Nice touches:

- The `pdaPath === undefined` fallback (PDA link not resolvable in the tree, e.g. purely external overrides) conservatively degrades to the old call shape rather than guessing.
- The `hasVariablePdaSeeds` guard incidentally fixes a pre-existing argument-misposition bug in the `pdaProgramValue` branch too (explicit `programId` account/arg override with no seed values would previously have landed the config in the seeds slot).
- Test coverage is thorough: same-program, other-program (via `rootNode`), pinned `programId`, argument defaults, and the empty-seeds positional case.

## Things to watch out for

- **`linkOverrides.pdas` contract change**: overridden finders must now accept a trailing `config: { programAddress?: Address }`. TypeScript consumers whose override finders declare a narrower signature will get compile errors on regenerate. The changeset documents this well, but see my inline question on the bump level.
- Generated output changes for any client with same-program linked PDA defaults — downstream clients (SPL token etc.) need regeneration, as noted in the PR description.

- CI enforces a clean working tree after build + tests, and `test/e2e/test.sh` regenerates projects. Only `anchor` and `token` were regenerated here — worth confirming the other e2e projects (`system`, `memo`, `dummy`) genuinely have no same-program linked PDA defaults so the clean-tree check passes (CI green should confirm this).
- The program identity comparison uses `programNode.name`, which is the linkable identity in the Codama tree — consistent with how `linkables` resolves links, so comparing by name rather than `publicKey` is the right call here.

.changeset/small-emus-happen.md
lorisleiva: minor is fine

lorisleiva (approved):
Thank you!

PR #192, opened 2026-08-20
Run E2E tests with Kit Plugins

lorisleiva:
This PR consolidates the E2E harness into the root package, generates clients through the Codama CLI, and replaces the local validator and custom LiteSVM utilities with official Kit Plugins whilst retaining AVA. It integrates the Node ESM compatibility fixture into the shared workflow, explicitly stabilises the shared tsup output layout, and updates the contributor guidance to describe the new pipeline.


lorisleiva:
* **#192** <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/192?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/192?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

trevor-cortex (approved):
This PR replaces the per-project E2E harness (individual `package.json`s, lockfiles, tsup/tsconfig configs, `test.sh`, `generate.cjs`, and a local `solana-test-validator`) with a single root-owned pipeline: generation via `codama run` configs, one shared tsup build into `test/e2e/dist`, one AVA config, and LiteSVM via the official Kit plugins (`@solana/kit-plugin-litesvm` / `@solana/kit-plugin-signer`). The test migrations are faithful — assertions are preserved or strengthened (e.g. exact balances in `transferSol` now that the fee payer is the plugin's funded payer, and the idempotent ATA test now actually invokes the instruction twice), and the LiteSVM-specific `expireBlockhash()` calls are commented where needed. Nice cleanup overall: the CI simplification (no Solana install, no nested `pnpm install`) and the deletion of ~16k lines of nested lockfiles are a clear win.

A few things worth noting:

1. **`CLAUDE.md` is now stale.** Its "Testing strategy → E2E tests" section still describes `test/e2e/test.sh`, `test/e2e/generate.cjs`, and starting `solana-test-validator` on port 8899 — all removed here. Since that file is the grounding for LLM contributors, it's worth updating it in this PR (or a fast follow) to describe the new `test:e2e` pipeline (`build → codama run → tsup/tsc → ava + node-esm checks`) and the LiteSVM plugin setup.
2. **`.prettierignore` narrowing**: now only `test/e2e/*/idl.json` is ignored, so the checked-in generated clients and the new test code are lint/prettier-checked at the root. That's a nice tightening, but it couples `pnpm lint` to the renderer's own Prettier formatting matching the root config — presumably CI green confirms this, just flagging it as the invariant to keep in mind when the renderer's formatting behavior changes.
3. **AVA glob ↔ tsup layout coupling**: `ava.config.cjs`'s `test/e2e/dist/*/test/*.test.js` glob relies on tsup's common-ancestor entry resolution producing `dist/<project>/test/` and `dist/<project>/src/` (which also makes the compiled tests' `../src/index.js` imports resolve to the bundled clients). It works, but it's implicit — if a future entry is added outside `test/e2e`, the common ancestor shifts and the layout silently breaks. A one-line comment in `tsup.config.ts` documenting this would help future maintainers.

No changeset needed since this is test/CI-only — correct per the repo's conventions. Approving; the inline comments below are minor questions/nits, none blocking.

trevor-cortex (approved):
Re-reviewed the latest changes since my previous approval. The delta addresses both of the follow-ups I raised:

1. **Contributor guidance updated** — `CLAUDE.md` and `AGENTS.md` now describe the actual pipeline: root-owned dependencies/build/test-runner, generation via each fixture's `codama.json` + Codama CLI, the shared tsup/tsc build, AVA against LiteSVM via the Kit plugins, and the separate `node-esm` NodeNext/type-stripping checks. The stale `test.sh` / `generate.cjs` / port-8899 references are gone, and the note that Codama configs load `dist/index.node.cjs` (so the root build must precede generation) is preserved. Both files were updated in lockstep, which matters since they're duplicates of each other.

2. **tsup output layout stabilised** — `esbuildOptions.outbase = './test/e2e'` now pins the output structure explicitly instead of relying on esbuild's common-ancestor inference. This is a stronger fix than the comment I suggested: the `dist/<project>/test/` and `dist/<project>/src/` layout that the AVA glob and the compiled tests' `../src/index.js` imports depend on can no longer silently shift if a future entry lands outside `test/e2e`.

The remaining inline nits from my first pass (exact-pinned `0.12.1` plugin versions, the potential double root build when CI runs `pnpm build` before `test:e2e` which itself runs `pnpm build`, and the `node-esm` codama config's `"."` + `generatedFolder` shape) were explicitly non-blocking and I won't re-raise them.

Still no changeset needed — test/CI/docs-only. Ship it.

PR #199, opened 2026-08-24
Migrate E2E tests from AVA to Vitest

lorisleiva:
This PR replaces AVA with Vitest as the runner for the shared E2E suite, completing the follow-up planned after #192. The tests now run as TypeScript directly through a dedicated Vitest config at `test/e2e/vitest.config.mts`, rooted at the E2E directory and kept separate from the root unit-test projects. All 19 active tests and the intentional Anchor skip are preserved with equivalent assertions. The tsup build keeps bundling each fixture's `src/index.ts` as a bundler-compatibility proof, and both `tsc` type-checks are unchanged, but the test files are no longer compiled to `dist` since Vitest executes them directly. The `node-esm` fixture deliberately stays on direct Node execution because it validates emitted ESM under NodeNext and Node's native type stripping, which a transform pipeline would bypass. The `ava` dependency, its config file, and all remaining references are removed.


lorisleiva:
* **#200** <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/200?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#199** <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/199?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/199?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #203, opened 2026-08-24
Add tests proving unset optional accounts preserve account ordering

### LINKED ISSUE #94: Optional accounts are not passed when None
When an optional account is None, it is not passed to the program. This makes it complicated to handle cases with multiple optional accounts or when optional accounts are not last in the list of accounts.

Passing the called program ID for None optional accounts is efficient in terms of tx size and CU, while it preserves the order of passed accounts.

lorisleiva:
This PR closes #94 by adding the test coverage that was missing for Codama's optional-account strategies, proving the reported problem does not occur under the default behaviour. Under the default `programId` strategy, an unset optional account is replaced by the program address rather than being dropped, so accounts keep their declared order even when an optional account precedes a required one; the legacy `omitted` strategy filters unset accounts out, matching Anchor's legacy behaviour. New unit tests in `instructionsPage.test.ts` assert the emitted code for both strategies, and a new `instruction11` in the `dummy` end-to-end fixture (an optional account followed by a required one) is exercised by build-and-parse round-trip tests confirming the placeholder preserves ordering and parses back to `undefined`. A separate test also documents that a `numberValueNode` renders correctly up to `Number.MAX_SAFE_INTEGER`, the largest integer it can carry exactly.


lorisleiva:
* **#203** <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/203?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #215, opened 2026-09-09
Widen the inputs accepted by generated instruction builders

lorisleiva:
This PR makes generated instruction builders accept any address-carrying value for their accounts by using the `InstructionAccountInput` and `InstructionSignerInput` types shipped in `@solana/kit` 8.3.0. Non-signer accounts accept an `Address`, a `ProgramDerivedAddress`, any object exposing an `address` property (which lets third-party wrappers such as web3.js's `PublicKey` be passed directly) or an account meta overriding the IDL-declared role, whilst signer accounts accept a `TransactionSigner` or an `AccountSignerMeta`. Builders declare one type parameter per account capturing the input provided for it — so the returned instruction's account metas reflect the exact values provided whilst excess property checks on the input are preserved — and forward each account's signer flag to `getAccountMetaFactory` so the runtime matches. Accounts whose bump seed is consumed by an argument are no longer required to be PDAs at the type level; providing a plain address for them throws at runtime unless the dependent argument is given explicitly. Explicit type arguments on builders are a breaking change for the rare call sites that use them, and generated code now requires `@solana/kit` `^8.3.0`.


lorisleiva:
* **#216** <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/216?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#215** <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/215?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/215?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

src/fragments/instructionFunction.ts
```diff
@@ -120,14 +120,25 @@ function getProgramAddressInitializationFragment(programAddressConstant: Fragmen
...
             const isWritable = account.isWritable ? 'true' : 'false';
-            return fragment`${name}: { value: input.${name} ?? null, isWritable: ${isWritable} }`;
+            // The resolved signer flag accounts for default values that may not be signers
+            // (e.g. a signer account defaulting to a PDA), in which case it becomes `'either'`.
+            const resolvedAccount = resolvedInputs.find(
+                (input): input is ResolvedInstructionAccount =>
+                    input.kind === 'instructionAccountNode' && input.name === account.name,
+            );
+            const resolvedIsSigner = resolvedAccount?.resolvedIsSigner ?? account.isSigner;
+            const isSigner = resolvedIsSigner === 'either' ? "'either'" : resolvedIsSigner ? 'true' : 'false';
```
trevor-cortex: Two things on the `resolvedIsSigner` sourcing:

1. **Untested branch.** This is the only non-mechanical logic added to the renderer and the unit test in `instructionsPage.test.ts` only covers `isSigner: 'either'` declared on the node itself. A small test with `isSigner: true` + a `pdaValueNode` (or `programIdValueNode`) default asserting `isSigner: 'either'` is emitted would pin the downgrade — that's the case where emitting `true` would make the builder throw `RESOLVED_INSTRUCTION_INPUT_MUST_BE_SIGNER` whenever the default kicks in, so it's the one that matters.

2. **Does the visitor ever go `false` → `'either'`?** The changeset promises that a `TransactionSigner` passed to a non-signer account merely carries its address. That holds only if `resolvedIsSigner` never relaxes an `isSigner: false` account (e.g. one defaulting to `payerValueNode`/`identityValueNode` or to a signer sibling via `accountValueNode`) to `'either'` — in that case the factory would upgrade the caller's signer after all. From memory `getResolvedInstructionInputsVisitor` only relaxes `true`, but worth a quick confirm since you own both sides; if it can happen, `account.isSigner === false ? 'false' : …` here would keep the runtime honest.

src/fragments/instructionFunction.ts
```diff
@@ -237,38 +248,56 @@ function getReturnTypeFragment(instructionTypeFragment: Fragment, hasByteDeltas:
...
+    programAddressConstant: Fragment,
+): Fragment {
+    const instructionNode = getLastNodeFromPath(scope.instructionPath);
+    const hasAccounts = (instructionNode.accounts ?? []).length > 0;
     return mergeFragments(
         [
-            ...(instructionNode.accounts ?? []).map(
-                account => fragment`TAccount${pascalCase(account.name)} extends string`,
-            ),
+            ...(hasAccounts ? [fragment`TInput extends ${getInputTypeName(scope)}`] : []),
```
trevor-cortex: Heads-up on a DX trade-off this shape introduces: with `input: TInput` (rather than a concrete `XInput<…>`), TypeScript infers `TInput` as the argument's own type, so fresh-object-literal excess property checking no longer applies. A misspelled *optional* account — e.g. `tokenProgam: TOKEN_2022_PROGRAM_ADDRESS` — now compiles and silently resolves to the IDL default. Required-account typos are still caught as missing keys.

If that's an accepted cost of brand inference (Kit's typetests validate this exact `TInput extends Input` shape, so I assume so), consider pinning it in `instructionInputs.test.ts` so it's a documented decision. If you'd rather keep the check, the standard exact-object trick works here:

```ts
input: TInput & { [K in Exclude<keyof TInput, keyof XInput>]?: never }
```

…at the cost of a noisier generated signature. Non-blocking either way.

trevor-cortex (approved):
## Summary

Adopts the Kit 8.3.0 (`anza-xyz/kit#2025`) program-client-core primitives in generated instruction builders:

- `instructionInputType.ts`: the five-way `Address` / `ProgramDerivedAddress` / `TransactionSigner` branch collapses to `InstructionAccountInput<T>` / `InstructionSignerInput<T>` (union of both for `'either'`). The `isPda` special-casing is gone, so bump-consuming accounts are no longer typed as PDAs.
- `instructionFunction.ts`: per-account `TAccountX extends string` type params are replaced by a single `TInput extends XInput`, and the return type is derived per account via `ResolvedInstructionAccountMeta<TInput['x'], InstructionAccountInputAddress<TInput['x']>[, SignerRole & AccountSignerMeta]>`. `originalAccounts` now carries `isSigner`, sourced from `resolvedIsSigner` so a signer account with a non-signer default degrades to `'either'`.
- Default `@solana/*` versions bumped to `^8.3.0`; root devDeps bumped; all e2e fixtures regenerated; a new `instruction12` fixture + `instructionInputs.test.ts` exercise the runtime matrix and pin the resolved meta types with `expectTypeOf`.

I cross-checked the emitted shapes against what landed in Kit 8.3.0: the `'either'` three-argument `ResolvedInstructionAccountMeta` form, `isSigner: 'either'` on `ResolvedInstructionAccount`, and the `TInput extends Input` / `input: TInput` builder shape are all what the Kit typetests validate, and the e2e `tsc` pass is the executable proof. Source changes are tight and follow the fragment conventions. Approving — nothing below is blocking, but the first item is a user-visible trade-off I'd like a conscious yes on.

## Things to watch out for

**1. Excess property checking on `input` is gone.** Previously `input: XInput<TAccountA, …>` was a concrete object type, so a misspelled *optional* account (`tokenProgam: TOKEN_2022_PROGRAM_ADDRESS`) was a compile error. With `input: TInput` and `TInput extends XInput`, TS infers `TInput` as the literal's own type — extra keys satisfy the constraint and the typo silently falls through to the IDL default (legacy Token program, in that example). Required-account typos are still caught (missing key), so it's only optional accounts, but those are exactly the ones with defaults. If this is an accepted cost of brand inference, I'd suggest pinning it with a `@ts-expect-error`-free "this compiles" case in `instructionInputs.test.ts` so the behaviour is deliberate rather than incidental; if not, the usual `TInput & { [K in Exclude<keyof TInput, keyof XInput>]?: never }` trick restores the check at the cost of noisier generated signatures. See inline at `instructionFunction.ts:266`.

**2. Accounts whose bump is consumed by an argument** move from a compile-time `ProgramDerivedAddress` requirement to a runtime `UNEXPECTED_RESOLVED_INSTRUCTION_INPUT_TYPE`. The changeset says so and the e2e covers both the throw and the explicit-`bump` escape. Just flagging it for anyone reading the diff of `getAccountTypeFragment` and wondering where `isPda` went.

**3. `resolvedIsSigner` sourcing** — see inline at `instructionFunction.ts:133-140`. The `'either'` downgrade is untested at the unit level and I want to confirm the visitor never *upgrades* `false`.

- **Bump level**: `minor` is consistent with this repo's precedent — `2.2.0`, `2.3.0` and `1.7.0` all raised the Kit floor as minors, and `2.4.0` shipped a generated-output change with "regenerate after upgrading". The explicit-type-argument break is real but rare, and the changeset calls it out. No objection.
- `remainingAccounts` inputs are still typed `Array<TransactionSigner | Address>` (see `getRemainingAccountsFragment` in `instructionInputType.ts` and the `typeof addressOrSigner === 'string'` runtime in `instructionRemainingAccounts.ts`). Widening those to `InstructionAccountInput` would need a runtime change too, so out of scope — but it's the one place a `PublicKey` wrapper still won't be accepted.
- The plugin surface (`XPluginInstructions`) uses `ReturnType<typeof getXInstruction>`, which instantiates `TInput` at its constraint and resolves every meta to the IDL default. Unchanged in spirit from before (the old `TAccountX` params also collapsed to `string`), so no regression.
- `README.md` doesn't document input types, so nothing to update there. `pnpm-lock.yaml` skipped.

test/e2e/dummy/test/instructionInputs.test.ts
```diff
@@ -0,0 +1,210 @@
...
+            // @ts-expect-error `bmup` is not a known input — excess property checks still apply.
+            bmup: 42,
+            delegate: delegateAddress,
+            metadata,
+            target: targetAddress,
+        });
+
+    // Then the misspelled input is ignored at runtime and the misspelling is a compile error.
+    expect(build().data).toStrictEqual(new Uint8Array([255]));
+});
```
trevor-cortex, line 128: Good — this pins the excess-property behaviour I asked about. One tiny optional follow-up: `instruction12` has no optional *accounts*, so `bmup` exercises the argument path. The case that actually bites users is a misspelled optional account silently falling back to its default (e.g. `tokenProgam`). `instruction11` has `optionalAccount?`, so a two-liner there with `// @ts-expect-error optinalAccount: …` plus an `expectTypeOf(ix.accounts[0]).toEqualTypeOf<ReadonlyAccount<string>>()` for the *omitted* case would cover both the typo and the "no inference site → constraint → IDL-declared meta" collapse in one go. Non-blocking.

trevor-cortex (approved):
## Re-review

The shape has changed materially since my first pass, so a fresh read rather than a delta:

- Builders and input types are back to **one type parameter per account** (`TAccountX extends InstructionAccountInput` / `InstructionSignerInput` / the union for `'either'`), with the input type itself now `{ x: TAccountX; … }` and defaults set to the constraint. The return type resolves each param through `ResolvedInstructionAccountMeta<TAccountX, InstructionAccountInputAddress<TAccountX>[, SignerRole & AccountSignerMeta]>`. Because `input: XInput<TAccountA, …>` is still a concrete object type after inference, excess property checking is preserved — which resolves my main concern from the first review, and the new `@ts-expect-error` case plus the "wider objects and spreads still compile" case pin both sides of that behaviour. Nice.
- `getRuntimeIsSignerFlag` now short-circuits `isSigner === false` to `'false'`, so a non-signer account can never be relaxed to `'either'` regardless of what the visitor resolves. Both directions are now covered at the unit level (`downgrades … whose default value may not be a signer` / `never upgrades …`). That closes my second point.
- The constraint fragment is shared between `instructionInputType.ts` and `instructionFunction.ts` via `getInstructionAccountInputConstraintFragment`, so the two can't drift. The docblocks on `getTypeParamsFragment` / `getInstructionTypeFragment` explain the *why*, which future readers of the generated output will appreciate.
- Changeset updated to describe the per-account type parameter semantics ("capture the input value provided for each account, rather than its address") and the explicit-type-argument break. Still `minor`, still consistent with repo precedent.

Both prior points addressed; approving.

- **Omitted optional accounts**: when an optional account isn't provided, its `TAccountX` has no inference site and falls back to the constraint (`InstructionAccountInput`). The plugin test (`ReturnType<typeof getX>` instantiates every param at its constraint) asserts that this collapses to `ReadonlyAccount<string>` / `WritableAccount<string>`, so Kit's `ResolvedInstructionAccountMeta` handles the full-union case correctly. The same path is exercised implicitly for `instruction11`'s `optionalAccount` via `tsc`, but there's no `expectTypeOf` on it — see inline for a tiny optional follow-up.
- Unchanged from my first review and still true: `remainingAccounts` inputs remain `Array<TransactionSigner | Address>` (runtime in `instructionRemainingAccounts.ts` does `typeof x === 'string'`), so a `PublicKey`-style wrapper still won't be accepted there. Out of scope, just don't be surprised.
- The Kit-side `TInput extends Input` shape I referenced in the first review is *not* what's used here anymore — this PR deliberately picked per-account params to keep excess-property checks. Don't flag the divergence from Kit's typetests as a problem; it's the point.

trevor-cortex (approved):
## Re-review (round 3)

The only change since my last pass is `test/e2e/dummy/test/instructionInputs.test.ts` (+26 lines); the renderer sources and generated fixtures are byte-for-byte what I reviewed last time.

The new `optional accounts resolve to the IDL-declared meta when omitted and reject misspellings` test covers exactly the follow-up I asked for on `instruction11`: the omitted `optionalAccount` is asserted to collapse to `ReadonlyAccount<string>` at the type level and to the program address at runtime, and the `optinalAccount` typo is pinned as a `@ts-expect-error` with a runtime check that it would otherwise have silently fallen back to the default. That's the case that actually bites users, so it's good to have it on record alongside the `instruction12` `bmup` case.

Nothing new to raise. Approving — ready to merge from my side.

Everything in my previous two reviews still applies (`remainingAccounts` not widened, per-account type params are a deliberate divergence from Kit's `TInput extends Input` typetest shape, `minor` bump consistent with repo precedent). No open threads.

mcintyre94 (approved):
Nice, these widened inputs are great!

lorisleiva:
### Merge activity

* **Sep 10, 4:28 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/codama-idl/renderers-js/215).
* **Sep 10, 4:29 PM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.com/github/pr/codama-idl/renderers-js/215).

PR #216, opened 2026-09-10
Accept the same inputs for remaining accounts as for instruction accounts

lorisleiva:
This PR completes the input widening started in #215 by having remaining accounts provided via the instruction input accept the same inputs as instruction accounts — `InstructionAccountInput`, `InstructionSignerInput` or the union of both depending on their signer flag — and by converting them with the same `getAccountMeta` helper, so that both share a single set of semantics: addresses are extracted from any address-carrying value such as a web3.js `PublicKey` wrapper, explicit account metas override the IDL-derived role, signers merely carry their address for non-signer remaining accounts, and values that cannot sign are rejected for signer remaining accounts. Remaining accounts backed by an existing instruction argument keep being typed by that argument since they are encoded in the instruction data.


lorisleiva:
* **#216** <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/216?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/216?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#215** <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/215?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

src/fragments/instructionFunction.ts
```diff
@@ -101,6 +101,7 @@ export function getInstructionFunctionFragment(
     const functionBody = mergeFragments(
         [
             getProgramAddressInitializationFragment(programAddressConstant),
+            getAccountMetaHelperFragment(instructionNode, hasAccounts || hasRemainingAccountArgs),
```
trevor-cortex: `hasRemainingAccountArgs` is true for *every* `argumentValueNode` remaining account, including ones whose argument already exists on the instruction — and those take the `AccountRole` branch in `getArgumentValueNodeFragment`, never calling `getAccountMeta`. With no instruction accounts, that leaves a `const getAccountMeta = getAccountMetaFactory(...)` that's never read (plus a spurious `getAccountMetaFactory` import), which is a `TS6133` under `noUnusedLocals` — a setting `test/e2e/tsconfig.json` itself enables; there just isn't a fixture with this shape.

The gate should only count remaining accounts that actually route through the helper, e.g.:

```ts
const allArgumentNames = new Set(getAllInstructionArguments(instructionNode).map(arg => arg.name));
const hasRemainingAccountInputs = (instructionNode.remainingAccounts ?? []).some(
    ({ value }) => isNode(value, 'argumentValueNode') && !allArgumentNames.has(value.name),
);
// ...
getAccountMetaHelperFragment(instructionNode, hasAccounts || hasRemainingAccountInputs),
```

Since this same "is backed by an existing argument" predicate already lives in `instructionInputType.ts` and `instructionRemainingAccounts.ts`, it might be worth extracting a small shared helper (e.g. `isRemainingAccountsBackedByArgument(instructionNode, node)`) and using it in all three spots.

test/instructionsPage.test.ts
```diff
@@ -1156,3 +1159,112 @@ test("it drops an unset optional account under the legacy 'omitted' strategy", a
...
+    // Then we expect the argument to keep its data type in the input.
+    await renderMapContains(renderMap, 'instructions/myInstruction.ts', [
+        "extraAccounts: MyInstructionInstructionDataArgs['extraAccounts'];",
+    ]);
+
+    // And we expect the remaining accounts to be derived from the IDL flags alone.
+    await renderMapContains(renderMap, 'instructions/myInstruction.ts', [
+        'const remainingAccounts: AccountMeta[] = args.extraAccounts.map( (address) => ({ address, role: AccountRole.READONLY_SIGNER }) );',
+    ]);
+    await renderMapDoesNotContain(renderMap, 'instructions/myInstruction.ts', ['getAccountMeta(']);
```
trevor-cortex: This assertion passes today for the wrong reason: the declaration `const getAccountMeta = getAccountMetaFactory(programAddress, 'programId');` *is* emitted for this instruction (see comment on `instructionFunction.ts`), it just doesn't match the substring `getAccountMeta(`. Once the gate is fixed, tightening this to `['getAccountMeta']` (or adding `'getAccountMetaFactory'`) would make the test actually guard against the unused helper.

trevor-cortex (changes requested):
## Summary

Completes the widening started in #215 for remaining accounts provided through the instruction input. Their input type now comes from the shared `getInstructionAccountInputConstraintFragment` (`InstructionAccountInput` / `InstructionSignerInput` / union for `'either'`), and they are converted with the same `getAccountMeta` helper as instruction accounts, wrapped in `getNonNullResolvedInstructionInput` to discharge the `undefined` branch. The helper declaration moves from the return statement to a dedicated `getAccountMetaHelperFragment` emitted near the top of the function body. Remaining accounts backed by an existing data/extra argument keep the old `Array<Address>` → `AccountRole` mapping. Generated fixtures, unit tests and E2E tests (dummy `instruction2`/`instruction8`, token `transfer`) are updated accordingly; changeset is `minor`, consistent with the repo's policy for user-facing widening.

The direction is right and the runtime semantics are well covered by the new E2E tests. One correctness issue needs fixing before merge.

## Needs a fix

**Unused `getAccountMeta` local in generated code.** `getAccountMetaHelperFragment` is gated on `hasAccounts || hasRemainingAccountArgs`, but `hasRemainingAccountArgs` counts every `argumentValueNode` remaining account — including those whose argument already exists on the instruction, which take the `AccountRole` branch and never call `getAccountMeta`. For an instruction with no accounts and only that kind of remaining accounts, the emitted function declares `const getAccountMeta = getAccountMetaFactory(...)` and never reads it, and drags in a spurious `getAccountMetaFactory` import. That's a `TS6133` for any consumer compiling with `noUnusedLocals` (which `test/e2e/tsconfig.json` itself enables — no fixture happens to have that shape, so CI doesn't catch it). The new unit test "backed by an instruction argument as plain addresses" is precisely that shape, but its `renderMapDoesNotContain([...'getAccountMeta('])` assertion passes only because the declaration reads `getAccountMeta = getAccountMetaFactory(`. Details inline.

## Notes for other reviewers

- The "does this remaining account map to an existing argument?" predicate is now needed in three places (`instructionInputType.ts`, `instructionRemainingAccounts.ts`, and — once fixed — `instructionFunction.ts`). Worth a tiny shared helper so they can't drift.
- Behaviour change to be aware of when reviewing downstream regenerations: `isSigner: false` remaining accounts previously only accepted `Address`; they now accept signers and silently carry just the address, mirroring instruction accounts. `isSigner: true` remaining accounts previously accepted `TransactionSigner` only; they now accept `InstructionSignerInput` with a runtime `RESOLVED_INSTRUCTION_INPUT_MUST_BE_SIGNER` guard. Both are covered by tests and match the intent, just flagging the widening.
- `getNonNullResolvedInstructionInput` throws with `inputName` set to the array's name, so a failing element isn't identified by index. Probably acceptable given `getAccountMeta` itself already reports that way, but it's a small UX difference from the old inline `map` code that could never throw.
- Optional nit: the new `@codama/nodes` imports in `test/instructionsPage.test.ts` (`prefixedCountNode`, `arrayTypeNode`, `instructionRemainingAccountsNode`) are inserted out of alphabetical order relative to the rest of that import list.

trevor-cortex (approved):
## Summary

Re-review after the follow-up push. The PR widens remaining accounts provided through the instruction input to accept the same `InstructionAccountInput` / `InstructionSignerInput` inputs as instruction accounts, and converts them via the shared `getAccountMeta` helper (now emitted near the top of the function body by `getAccountMetaHelperFragment`), wrapped in `getNonNullResolvedInstructionInput`. Remaining accounts backed by an existing data argument keep the `Array<Address>` → `AccountRole` mapping.

## Prior feedback — resolved

**Unused `getAccountMeta` local / spurious import** — fixed. The helper is now gated on `hasAccounts || hasRemainingAccountInputs(instructionNode)`, where `hasRemainingAccountInputs` in the new `src/utils/remainingAccounts.ts` explicitly excludes argument-backed remaining accounts. The predicate is centralised in `isRemainingAccountsBackedByArgument` and used consistently in `instructionInputType.ts`, `instructionRemainingAccounts.ts`, and (via `hasRemainingAccountInputs`) `instructionFunction.ts`, so the three call sites can no longer drift. The coverage is also stronger than what I asked for:

- The unit test now asserts `renderMapDoesNotContain(..., ['getAccountMeta'])` (bare name), which catches both the declaration and the `getAccountMetaFactory` import rather than just the call.
- The new `instruction13` dummy fixture is precisely the shape that would previously have produced a `TS6133` under `noUnusedLocals` (no accounts, only an argument-backed remaining account), and it's compiled as part of the E2E suite. The generated output correctly has no helper and no `getAccountMetaFactory` import.

Small correction on my side: the `@codama/nodes` import ordering nit in `test/instructionsPage.test.ts` was wrong — `arrayTypeNode`, `instructionRemainingAccountsNode` and `prefixedCountNode` are all in the right alphabetical positions. Ignore that one.

## Notes for other reviewers

- Body ordering in the generated function is `programAddress` → `getAccountMeta` helper → `originalAccounts`/`accounts` → `args` → resolver scope → resolved inputs / remaining accounts → return. The helper only depends on `programAddress`, so hoisting it above the accounts block is safe; the generated fixtures (e.g. `transferSol.ts`, `transfer.ts`) confirm the move.
- `hasRemainingAccountArgs` in `instructionFunction.ts` still counts *all* `argumentValueNode` remaining accounts. That's intentional — it feeds `hasAnyArgs`, and argument-backed remaining accounts still read from `args`, so `const args = { ...input }` must be emitted for them too. Only the helper gating uses the narrower predicate.
- The behaviour widening I flagged last time still applies and is now well covered: `isSigner: false` remaining accounts accept signers and carry only their address (`instructionInputs.test.ts`), `'either'` upgrades to a signer meta only when a signer is passed (`instruction8`), and `isSigner: true` rejects non-signers with `RESOLVED_INSTRUCTION_INPUT_MUST_BE_SIGNER` both at the type level (`@ts-expect-error`) and at runtime (`transfer.test.ts`).
- `minor` changeset is consistent with #215 and the repo's policy for user-facing input widening.

mcintyre94 (approved):
Nice!

lorisleiva:
### Merge activity

* **Sep 10, 4:28 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/codama-idl/renderers-js/216).
* **Sep 10, 4:30 PM UTC**: [Graphite](https://app.graphite.com/github/pr/codama-idl/renderers-js/216) rebased this pull request as part of a merge.
* **Sep 10, 4:30 PM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.com/github/pr/codama-idl/renderers-js/216).

PR #217, opened 2026-09-10
Fix the `preferRoot` import strategy leaking into subsequent renders

lorisleiva:
This PR fixes a shared-state bug in the import map resolution: the `preferRoot` kit import strategy mutated the module-level default module map in place, so that any later render using the `rootOnly` strategy within the same process would import program client helpers from `@solana/program-client-core` instead of `@solana/kit/program-client-core`. The default module map is now built fresh for each resolution, and the underlying constants are frozen and typed as read-only so that such a mutation can no longer compile. A regression test resolving the same import map with `preferRoot` and then `rootOnly` covers the fix.


lorisleiva:
* **#217** <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/217?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/renderers-js/217?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

src/utils/importMap.ts
```diff
@@ -200,6 +194,22 @@ function resolveImportMapModules(
...
+function getDefaultExternalModuleMap(kitImportStrategy: KitImportStrategy): Record<string, string> {
+    switch (kitImportStrategy) {
+        case 'granular':
+            return { ...DEFAULT_GRANULAR_EXTERNAL_MODULE_MAP };
+        case 'preferRoot':
+            return { ...DEFAULT_EXTERNAL_MODULE_MAP, solanaProgramClientCore: '@solana/program-client-core' };
+        case 'rootOnly':
+            return { ...DEFAULT_EXTERNAL_MODULE_MAP };
+    }
+}
```
trevor-cortex, line 211: Optional nit: the only caller immediately spreads this return value into `dependencyMapWithDefaults`, so the `{ ...DEFAULT_GRANULAR_EXTERNAL_MODULE_MAP }` / `{ ...DEFAULT_EXTERNAL_MODULE_MAP }` copies in the `granular` and `rootOnly` branches are technically redundant — you could return the frozen constants directly with a `Readonly<Record<string, string>>` return type and rely on the type + `Object.freeze` for safety.

That said, the defensive copy is cheap and makes the helper safe for any future caller that *doesn't* spread, which the docblock explicitly promises. Happy either way; just flagging it so the choice is deliberate.

trevor-cortex (approved):
## Summary

Fixes a shared-state bug in `resolveImportMapModules`: the `preferRoot` branch wrote `solanaProgramClientCore` directly onto the module-level `DEFAULT_EXTERNAL_MODULE_MAP`, so once any `preferRoot` resolution had run in a process, every subsequent `rootOnly` resolution (same map object) would emit `@solana/program-client-core` instead of `@solana/kit/program-client-core`. The fix extracts a `getDefaultExternalModuleMap(strategy)` helper that returns a fresh object per call, freezes both default maps and types them as `Readonly<Record<string, string>>`, and adds two tests: one asserting the resolved module per strategy, and one specifically reproducing the `preferRoot` → `rootOnly` leak.

## Assessment

- The root cause analysis is right and the fix is the correct shape. `Readonly<...>` makes the old assignment a compile error, and `Object.freeze` makes it a runtime `TypeError` in strict-mode modules, so this class of bug is closed off on both fronts.
- The `switch` over `KitImportStrategy` is exhaustive with no `default`, so adding a fourth strategy later will fail type-checking until it's handled here. That's a good property to have for this function.
- Both call sites of `resolveImportMapModules` (`importMapToString` and `getExternalDependencies`, which feeds `packageJson.ts` dependency syncing) go through the new helper, so the package.json sync path is fixed as well, not just the emitted imports.
- Test coverage is appropriate. The dedicated leak test exercises the exact failing order (`preferRoot` first, then `rootOnly`) on the same `ImportMap` instance, which is what would have caught the original bug.
- Changeset is `patch`, which matches a bug fix with no API change.

One optional nit inline about a redundant copy; not blocking.

Nothing else in the repo references the two default maps (they're not exported), so there's no other mutation path to audit. The E2E fixtures each run in their own Codama CLI invocation so they wouldn't have been affected by the leak; the bug only manifested when multiple renders with different strategies ran in one process (e.g. the unit test suite or a programmatic multi-client build).

